Issue #11, opened 2019-08-29
storybook does not handle react-native@0.61.0 new fast refresh mode


mgcrea:
**Describe the bug**

Following the release of react-native@0.61.0-rc.0, legacy "hot module reloading" and "live reloading" have been merged in a single new "fast refresh" mode to hot reload components (more details in this [blog post](https://medium.com/@jonnyburger/first-look-react-native-0-61-with-fast-reloading-ad387e502e6f)).

Unfortunately, this new mode does not work with storybook as I get ` WARN  Story with id button--default-view already exists in the store!` errors when fast refresh kicks in.

Since livereload isn't an option anymore, it makes developing components with storybook highly impractical as you don't have any refresh anymore.

**To Reproduce**
1. Create a new react-native project with `react-native init AwesomeProject --version react-native@next`
2. Install storybook, and save/update a story
3. See warning, no refresh

**Expected behavior**
Working fast refresh while editing a story

**System:**
```
Environment Info:

  System:
    OS: macOS 10.14.5
    CPU: (12) x64 Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
  Binaries:
    Node: 10.16.0 - ~/.nvm/versions/node/v10.16.0/bin/node
    Yarn: 1.17.3 - /usr/local/bin/yarn
    npm: 6.9.0 - ~/.nvm/versions/node/v10.16.0/bin/npm
  Browsers:
    Chrome: 76.0.3809.132
    Firefox: 68.0.1
    Safari: 12.1.1
  npmPackages:
    @storybook/react-native: ^5.1.11 => 5.1.11 
```

Note that I'm using the following hack to have working hooks:

```
addDecorator(Story => <Story />);
```

Gongreg:
Hello. I've managed to get fast refresh to work (no warnings), and actually adding/removing stories works.

There is one issue though, we will have to add a limit of calling only one storiesOf per file.

I would like your opinion regarding it, will it be an issue?

If not I would rather keep it like that, since we will try to also add CSF format in the future.

The technical side:

[...]

PR #30, opened 2020-02-15
Fix Fast refresh

Gongreg:
Issue: #11

## What I did

Updated getStorybookUI to reuse channel on rerender.

## TODO

* Update story list to not jump around on fast refresh.
* Make a PR to main storybook to fix ClientAPI.

## How to test

- Still manual testing for cases like these.

### DIFF
--- a/app/react-native/src/preview/index.tsx
+++ b/app/react-native/src/preview/index.tsx
@@ -54,7 +54,10 @@ export default class Preview {
     this._addons = {};
     this._decorators = [];
     this._stories = new StoryStore({ channel: null });
-    this._clientApi = new ClientApi({ storyStore: this._stories, disableAddStoryHotReload: true });
+    this._clientApi = new ClientApi({
+      storyStore: this._stories,
+      noStoryModuleAddMethodHotDispose: true,
+    });
   }
 
   api = () => {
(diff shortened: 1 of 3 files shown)


Gongreg:
@shilman, @ndelangen could you remove now bot from RN repo?

ndelangen:
I removed now.sh

dannyhw:
rebased this with master to reevaluate the status of this pr

dannyhw:
after looking through old issues it seems like for this to work storiesOf should be called once per file see #31.

ndelangen:
yes, storiesOf should be once per file, that's how it's also defined for CSF

dannyhw:
this change relies on https://github.com/storybookjs/storybook/pull/9868 and so we would need this change for 5.3.x in order to fix this for this version of react native storybook

dannyhw:
https://github.com/storybookjs/storybook/pull/9868 has now been released for version 5.3 with the 5.3.21 release 🥳. Should now be able to give this one another go.

dannyhw:
I got a bit stuck with this because I was finding it really hard to reproduce in the issue in the example project. Thinking this one might be delayed for the next release.

mohsinulhaq:
@Gongreg any issues in this? would be really helpful if this was fixed :)

dannyhw:
@mohsinulhaq I had issues with testing this change which makes it hard to tell if these changes work or not otherwise I would have merged it already.

Also there will be significant changes to this part of the code to support 6.0 so it might be easier fixed in 6.0.

mohsinulhaq:
@dannyhw No problem. Thanks!

dannyhw:
Will be fixed as part of 6.0 CSF support etc.

PR #74, opened 2020-07-23
fix(): update react-native-modal-selector to new version

jon-moreira:
Issue:

## What I did
Just update the external dependency ```react-native-modal-selector``` inside of ```ondevice-knobs``` addons to prevent error bellow
Image from iOS

## How to test

- Is this testable with Jest or Chromatic screenshots? no
- Does this need a new example in the kitchen sink apps? no
- Does this need an update to the documentation? no


dannyhw:
fixes #62

PR #76, opened 2020-08-03
[documentation] add readme from old project location

dannyhw:
## What I did

Copied over the readme from the old location of react native storybook

## How to test

Not testable, however perhaps you can check that the steps included in the docs actually work.


dannyhw:
I figure that having the old version of the readme would atleast be more useful than the todo message :).

PR #78, opened 2020-08-03
fix: attempt to resolve issue with dangerjs

dannyhw:
Issue:

Not sure if this will work, just creating this PR to see if it changes anything with the dangerjs check

## What I did

use yarn instead of npm in the dangerjs file

## How to test

see if checks pass


dannyhw:
in the latest run it seemed like there was an issue with the access to github, maybe the api token is wrong?

Issue #67, opened 2020-07-20
When viewing stories on Android they clash with the status bar


skywickenden:
**Describe the bug**
When viewing stories on Android they clash with the status bar

**To Reproduce**
1. View a react-native storybook on Android 

**Expected behavior**
The stories and storybook navigation should appear below the status bar.

**Screenshots**
![Screenshot_20200720-145303](https://user-images.githubusercontent.com/4930551/87946885-8302e200-ca9a-11ea-9f88-9e34165bebf0.png)

**System:**
(Using EXPO to publish to android from a mac)
  System:
    OS: macOS 10.15.3
    CPU: (8) x64 Intel(R) Core(TM) i7-8569U CPU @ 2.80GHz
  Binaries:
    Node: 12.16.2 - ~/.nvm/versions/node/v12.16.2/bin/node
    Yarn: 1.19.1 - /usr/local/bin/yarn
    npm: 6.14.4 - ~/.nvm/versions/node/v12.16.2/bin/npm
  Browsers:
    Chrome: 84.0.4147.89
    Safari: 13.0.5
  npmPackages:
    @storybook/addon-actions: ^5.3.19 => 5.3.19 
    @storybook/addon-knobs: ^5.3.19 => 5.3.19 
    @storybook/addon-links: ^5.3.19 => 5.3.19 
    @storybook/addon-ondevice-knobs: ^5.3.19 => 5.3.19 
    @storybook/addons: ^5.3.19 => 5.3.19 
    @storybook/react: ^5.3.19 => 5.3.19 
    @storybook/react-native: ^5.3.19 => 5.3.19 

**Additional context**
This is handled in iOS through the use of `SafeAreaView` in [OnDeviceUI](https://github.com/storybookjs/react-native/blob/master/app/react-native/src/preview/components/OnDeviceUI/index.tsx#129)

To solve this on Android we could create a SafeView that also takes care of Android. 

```
import React from 'react';
import { SafeAreaView, StyleSheet, Platform, StatusBar } from 'react-native';

const styles = StyleSheet.create({
  AndroidSafeArea: {
    paddingTop: Platform.OS === 'android' ? StatusBar.currentHeight : 0
  }
});

const SafeView = ({ children }) => <SafeAreaView style={styles.AndroidSafeArea}>{children}</SafeAreaView>;
export default SafeView;
```

skywickenden:
Would a PR be welcome?

dannyhw:
[...]

PR #88, opened 2020-08-07
fix: potential fix for status bar covering stories on android+expo

dannyhw:
Issue: #67 

## What I did

added a padding top for the safe area view when using android, this needs testing for both expo android and regular native android


dannyhw:
Needs more work because the issue seems to be specific to android with expo, native android didn't have this problem.

dannyhw:
I've added a check for expo (not sure about this) that seems to work. Should check this with expo 39 also

ndelangen (approved):

Issue #114, opened 2020-10-07
Themed values from emotion styled components are not working on on-device knobs


dannyhw:
**Describe the bug**

Currently the theme values are not being used in the knobs plugin on the device. This caused all the border colors to become much darker and certain UI changes since 5.3.22

**To Reproduce**
1. change a value in the theme
2. knobs such as text don't respect the border color for example

**Expected behavior**

Knobs should correctly use colors from the theme

example:

const StorybookUIRoot = getStorybookUI({ theme: { borderColor: 'red', } });

PR #125, opened 2020-10-30
feat: custom theme implementation

dannyhw:
Issue: #114 

## What I did

Due to issues with getting the types for emotion to work and just in general struggling to get the theme values to work across the different packages I made a simple implementation of some of the features that were being used from emotion. 

This introduces a new package that I've called @storybook/ondevice-theme which mimicks some features of emotion however I've only implemented explicitly what is being used in the project at this point. For example I didn't implement a styled component for flatlist since it's not used anywhere. Also These theming changes only effect ondevice ui and the server is not included.

With this change the theme values should now be respected in addons and in the base package and properly change when adjusted by the user. The way the user interacts with the theme through getstorybookUI works the same way by passing the theme prop.

**Note that there is an issue with hooks currently in the examples so I've left some TODO notes with improvements I intend to make**

Also the way the dependency works I need to test, hopefully this won't require the user to install the ondevice-theme package and it will be more of an internal dependency (that was my intention).

## How to test

Firstly just running the example app you should notice that some of the styles that were broken by a previous release are now back to normal (harsh black borders now gone).

Then you can experiment with changing some of the theme values in the example app.

Heres all the possible theme values:

```
const StorybookUIRoot = getStorybookUI({
  asyncStorage: false,
  theme: {
      backgroundColor: 'white',
      headerTextColor: 'black',
      labelColor: 'black',
      borderColor: '#e6e6e6',
      previewBorderColor: '#b3b3b3',
      buttonTextColor: '#999999',
      inputBorderColor: '#DDDDDD',
      buttonActiveTextColor: '#444444',
      positive: '#66BF3C',
      seconday: '#1EA7FD',
      primary: '#FF4785',
  }
});
```


dannyhw:
as a side note I also fixed some missing padding in the storylist where the story name had no horizontal padding

dannyhw:
Recently emotion 11 was released so I should try using that to see if it resolves the issues here.

dannyhw:
Closing this PR for now since its a bit dead

Issue #132, opened 2020-11-13
number knob ignores comma as decimal separator


sgonzalez-orangeloops:
**Describe the bug**
The number knob ignores numbers written after a comma. This is not desired when the device is configured with certain regions, like Russia, because in those cases iOS will pop up a keyboard with no option to write a dot.

**To Reproduce**
1. Set the region of a iOS device to russia (no need to change the language).
2. Add a number knob to a story.
3. Try to write number with decimals in the knob input.

**Expected behavior**
The number knob text input recognizes both the comma and the dot as decimal separators.

**Screenshots**
![ezgif-3-f2bbea0cc193](https://user-images.githubusercontent.com/56684338/99079419-ae0bc700-259e-11eb-97fa-76b41539dc4c.gif)

**System:**
```
System:
  OS: macOS 10.15.7
  CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
Binaries:
  Node: 15.0.1 - /usr/local/bin/node
  npm: 6.14.8 - /usr/local/bin/npm
Browsers:
  Chrome: 86.0.4240.193
  Edge: 86.0.622.63
  Safari: 14.0
npmPackages:
  @storybook/addon-actions: 5.3.18 => 5.3.18 
  @storybook/addon-knobs: 5.3.18 => 5.3.18 
  @storybook/addon-ondevice-actions: 5.3.18 => 5.3.18 
  @storybook/addon-ondevice-knobs: 5.3.23 => 5.3.23 
  @storybook/addons: 5.3.18 => 5.3.18 
  @storybook/react-native: 5.3.23 => 5.3.23 
  @storybook/react-native-server: 5.3.18 => 5.3.18
```

sgonzalez-orangeloops:
Found [this commit](https://github.com/storybookjs/react-native/commit/1e787fa5bcd2e4504cdf7a0cb8f0b14382062558) that removes logic that might be helpful in this case, but I couldn't find context for that removal.

dannyhw:
Was released as part of ondevice-knobs v5.3.24

PR #136, opened 2020-11-17
Convert coma to dot if it's the last non-digit char

sangonz193:
Issue: #132

## What I did
Modifies the number ondevice-knob so that it replaces a comma with a dot if it's the last non-digit char.

### DIFF
--- a/addons/ondevice-knobs/src/types/Number.js
+++ b/addons/ondevice-knobs/src/types/Number.js
@@ -15,45 +15,45 @@ const Input = styled.TextInput(({ theme }) => ({
 class NumberType extends React.Component {
   constructor(props) {
     super(props);
-    this.renderNormal = this.renderNormal.bind(this);
-    this.renderRange = this.renderRange.bind(this);
 
     const { knob } = this.props;
     const initialInputValue = Number.isNaN(knob.value) ? '' : knob.value.toString();
-    this.lastInputValue = initialInputValue;
+
     this.state = {
       inputValue: initialInputValue,
+      showError: false,
     };
   }
 
-  componentDidUpdate() {
-    const { onChange } = this.props;
-    const { inputValue } = this.state;
-
-    if (this.lastInputValue !== inputValue) {
-      const inputValueEmpty = inputValue.trim() === '';
-      const parsedInputValue = inputValueEmpty ? NaN : Number(inputValue);
-
-      if (!Number.isNaN(parsedInputValue) || inputValueEmpty) {
-        onChange(parsedInputValue);
-        this.lastInputValue = inputValue;
-      }
-    }
-  }
-
-  onChangeNormal = (value) => {
-    this.setState({
-      inputValue: value.replace(/,(?=\d+$)/, '.'),
-    });
-  };
-
-  renderNormal() {
+  shouldComponentUpdate(nextProps, nextState) {
     const { knob } = this.props;
     const { inputValue } = this.state;
-    const showError =
-      inputValue.trim() !== '' &&
-      (Number(inputValue).toString() !== knob.value.toString() || Number.isNaN(knob.value));
 
+    return nextProps.knob.value !== knob.value || nextState.inputValue !== inputValue;
+  }
+
+  onChangeNormal = (text) => {
+    const { onChange } = this.props;
+    const inputValue = text.trim().replace(/,/, '.');
+    if (inputValue === '') {
+      this.setState({ showError: false, inputValue });
+      return;
+    }
+
+    const parsedValue = Number(inputValue);
+
+    this.setState({ inputValue });
+
+    if (!Number.isNaN(parsedValue)) {
+      onChange(parsedValue);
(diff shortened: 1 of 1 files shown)


dannyhw:
Thanks for your contribution, will test it soon :)

dannyhw:
It seems to work for a comma but if you use a "." it just deletes it.

sangonz193:
Hmm I could make it work with a "." like this:
screenshot

Was your try any different? The regex I added should't change anything if there isn't any comma.

dannyhw:
delete the "." and try to add it again

I'll try again maybe I did something weird.

dannyhw:
ok actually what is happening is that the value is sometimes updated slowly and it deletes stuff after I type it.
I'll make a gif or something to show it in a sec

dannyhw:
weird2

dannyhw:
I was playing around with it and this is something that worked for me

remove the nan check since it doesn't seem to do much
```
onChangeNormal = (value) => {
    const { onChange } = this.props;
    const number = value.replace(/,(?=\d+$)/, '.');
    onChange(number);
  };
```

I think this will work, remove the transformer and make it so the value is not controlled directly.
```
<Input
        autoCapitalize="none"
        underlineColorAndroid="transparent"
        defaultValue={(knob.value || '').toString()}
        keyboardType="numeric"
        onChangeText={this.onChangeNormal}
      />
```

What do you think?

sangonz193:
That actually makes a lot of sense to me, especially the value -> defaultValue change.
I'll push the changes soon.

addons/ondevice-knobs/src/types/Number.js
```diff
@@ -17,33 +17,24 @@ class NumberType extends React.Component {
...
-
-  numberTransformer = (x) => {
-    if (Number.isNaN(Number(x))) {
-      return x.substr(0, x.length - 1);
-    }
 
-    return x;
-  };
+    const { knob } = this.props;
+    this.defaultValue = knob.value;
```
sangonz193: I wasn't expecting it to behave this way but for some reason if you keep changing the TextInput's defaultValue it'll just keep updating like if it was the value prop. That's why I added this, so that defaultValue doesn't change.

dannyhw: its probably because the parent component is being re-rendered when it shouldn't

sangonz193: I don't think preventing renders is the way to go in this case. I'd like to keep this approach unless you suggest something else.

dannyhw: Yeah anyway long term this addon will be deprecated to make way for controls in 6.0. Minimal changes to get things working is fine.

sangonz193:
Is there anything else I can do about this?

dannyhw:
@sangonz193 hey really sorry, been busy :/ will check this again tonight.

dannyhw:
@sangonz193 hey so I tested it and there is some other issues since now the value doesn't get updated by knob.value then the value that gets output is different than the one that you see. 

For example in the KnobsExample story the number might be 12.98 but the text in the input can show 12a,98. I think the problem with numbers being deleted is unavoidable if we control the value and having them not be synced is weird too.

My opinion is that we shouldn't filter the number at all and just warn the user that the value is not a number, like with a red border for example. What do you think?

sangonz193:
I like notifying the user there's something wrong with their input. Do you know of any knob that reports errors already? If there aren't, a red border sounds good for me.

Are there any considerations I follow to add the red color? any specific red? use/add a variable somewhere?

dannyhw:
@sangonz193 the design system for storybook isn't really enforced here but it probably makes sense to use the colors from there https://5ccbc373887ca40020446347-vehbzpmjqt.chromatic.com/?path=/story/design-system-colors--page

The error color I think is theme.color.negative which is #FF4400. I don't think any other knobs have this behaviour but if you're up to it then try it out for this input and then we can see if it fits well.

sangonz193:
I made the input component controlled again but added a state to handle the value. The idea is to show the red border only when the value `knob.value` represents is different from the one the state represents (unless we are talking about an empty input, in which case it's treated as a valid NaN value).

dannyhw:
@sangonz193 ok I will check when I can, thanks again for your contribution :)

dannyhw:
@sangonz193 I hope you don't mind, I made some small alterations to simplify the code a bit, can you test it a bit and let me know what you think? If you're happy with it I think we can merge this.

sangonz193:
Sorry for the delay. I tested your changes and they work excellent!

dannyhw:
@sangonz193 hey, great I will merge and I plan to make an alpha with some of the most recent bugfixes sometime this week.

dannyhw:
@sangonz193 you can test this by installing the new alpha versions
"@storybook/addon-ondevice-knobs": "5.3.24-alpha.1",
"@storybook/react-native": "5.3.24-alpha.1",

PR #165, opened 2021-05-08
Update README.md

wolverineks:
Issue:

## What I did

added async storage to parameters in readme


dannyhw:
thanks for your contribution, however it would be better if you could include a description in your PR

Issue #170, opened 2021-05-11
[6.0] add react native web to the example app


dannyhw:
As part of the work by @oblador to remove no-hoisting we removed some of the example projects to slim down the project. This meant removing the expo example which had a web example.

We should add react-native-web to the example app and get something basic rendering.

There may be some changes needed in the project to make this work. Making this work with ondevice knobs will require #120 to be fixed so that should be considered out of the scope of this issue

dannyhw:
Try https://aureliomerenda.medium.com/create-a-native-web-app-with-react-native-web-419acac86b82

If that doesn't work either change the example project to an expo one or you can try doing a more involved setup with custom webpack config.

Here is a guide with more custom webpack stuff that might be useful also (not tested this)
https://levelup.gitconnected.com/react-native-typescript-and-react-native-web-an-arduous-but-rewarding-journey-8f46090ca56b

Issue #162, opened 2021-05-06
[6.0] Change the workspace to use hoisting


dannyhw:
Currently in the monorepo no-hoisting is being used in the examples and throughout the project. This is generally used to make react native work in mono repo, however it would be better to hoist dependencies to improve install time and reduce the size of the repo.

This could also fix issues with hooks in the repo.

PR #172, opened 2021-05-12
chore: Remove expo projects and remove no-hoisting

oblador:
Issue: #162

## What I did

In order to make the monorepo a bit more manageable I've removed `no-hoisting` directives so that all packages can share the same installation of react and react-native. Since expo uses their own version of react-native and thus wouldn't work with hoisting, these examples have been removed as they provided little value anyways. Expo brought react-native-web setup, but we'll be adding that ourselves directly instead as tracked in #170.

Fresh installs are now ~8x faster and react hooks work. 

I've disabled Flipper on iOS as it doesn't work out of the box right now and makes compile time much slower. We can easily re-enable it later. 

I chose to keep using the 0.63 range as 0.64 brings react 17, an upgrade I prefer to do in a separate PR. 

## How to test

Run the example project as before. Minor caveat is that you must start metro manually.


dannyhw (commented):
Looks good, I had an issue when running android though, could you check?

```

FAILURE: Build failed with an exception.

* What went wrong:
Could not initialize class org.codehaus.groovy.runtime.InvokerHelper
```

__

after all it was a problem with my java_home variable. Looks good to me 👍

dannyhw (approved):
🥇

dannyhw:
thanks for your contribution 🙏

Issue #159, opened 2021-05-03
In Expo 41 the app never loads due to promise not being fulfilled


dannyhw:
**Describe the bug**

Due to #20 the expo loading screen never goes away probably because it's waiting for a promise to resolve.

**To Reproduce**
1. expo init APPNAME
2. npx -p @storybook/cli sb init --type react_native
3. add `export {default} from "./storybook"` in app.tsx
4. app never loads

**Expected behavior**
App should load and promises should resolve correctly

**Additional context**
If you comment out line 11 in node_modules/@storybook/addons/dist/index.js you can see that the issue stops 
`//require("core-js/modules/es.promise"); `

This requires a more permanent solution of course but this illustrates the problem.

dannyhw:
I've been talking with the storybook maintainers about a possible solution to this. Ideally there would be a modern build of storybook/addons available or atleast one without the es.promise polyfill. 

Recently i.e11 support was separated out into a separate repo so this could make it possible to provide a solution without the promise polyfill since ie.11 is the only browser without support for promises.

dannyhw:
A [PR](https://github.com/storybookjs/storybook/pull/14954) was created on the main storybook repository with a fix for future version of storybook. This means that this issue will be resolved in 6.0 of react native storybook. The PR enables provides a modern version of the distributed files, in your project you would add the following config `resolverMainFields:['sbmodern', 'main']` and the storybook dependencies used would be without polyfils.

nickplee:
A workaround for Storybook 5.3 that worked for me, assuming you're ok with another dependency, was to add the following to my app's `index.js`:

```javascript
import Bluebird from 'bluebird';
global.Promise = Bluebird;
```

grahamlyus:
Thanks for this. I just encountered the same issue in a bare react-native app using expo-font.

Issue #55, opened 2020-05-23
React Native server stories list infinitely loading


bouncycatt:
Hello! I had a problem with loading Storybook web explorer, it doesn't load the left menu. But on a device Storybook works well. I met the problem both during Intro React Native Storybook Tutorial and in my bare workflow project. It just doesn't build a web version. 

**Steps**
1. In a terminal to a project folder:
`npx -p @storybook/cli sb init --type react_native`
`yarn add -D @storybook/addon-ondevice-actions`

2. In storybook/rn-addons.js
`import '@storybook/addon-ondevice-actions/register';`

3. In the terminal I tried this and reverse order:
`yarn storybook`
`yarn start`
`a` – both Android simulator and a real Android device, Storybook works well on both options but does not in a browser.

4. On `yarn storybook` the answer:
```
yarn run v1.22.4
$ start-storybook -p 7007
info => Loading presets
info => Loading custom manager config.
╭────────────────────────────────────────────────────╮
│                                                    │
│   Storybook 5.3.18 started                         │
│   4.39 s                                           │
│                                                    │
│    Local:            http://localhost:7007/        │
│    On your network:  http://192.168.0.104:7007/    │
│                                                    │
╰────────────────────────────────────────────────────╯
```

**From the tutorial it should be this**
<img width="356" alt="from tutorial" src="https://user-images.githubusercontent.com/63786160/82722658-62ecb780-9cd1-11ea-97ad-5e7a3d04ff3c.png">

**But I have this result on any projects and environments I tried**
<img width="843" alt="error" src="https://user-images.githubusercontent.com/63786160/82723362-12785880-9cd7-11ea-87e9-a2142f46cbe9.png">

**Environment Info:**
  System:
    OS: macOS Mojave 10.14.6
    CPU: (8) x64 Intel(R) Core(TM) i7-4980HQ CPU @ 2.80GHz
  Binaries:
    Node: 12.16.3 - /usr/local/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
  Browsers:
    Chrome: 81.0.4044.138
    Firefox: 76.0
    Safari: 12.1.2
  npmPackages:
    @storybook/addon-actions: ^5.3.18 => 5.3.18 
    @storybook/addon-links: ^5.3.18 => 5.3.18 
    @storybook/addon-ondevice-actions: ^5.3.18 => 5.3.18 
    @storybook/addons: ^5.3.18 => 5.3.18 
[...]

PR #174, opened 2021-05-18
fix: use modern build of storybook to fix unresolved promises and other issues caused by polyfills [6.0]

### LINKED ISSUE #20: Promise.finally is being overwritten in React Native
**Describe the bug**
The inclusion of Storybook inside a React Native app seems to remove the implementation of `Promise.prototype.finally`. Causing the `Promise.resolve(...).finally is not a function error` to occur.

**To Reproduce**
1. Create a fresh React Native project.
2. Setup Storybook as reference here, https://storybook.js.org/docs/guides/guide-react-native/ `npx -p @storybook/cli sb init --type react_native` (no server)
3. Add the code snippet reference below and you’ll see initially you can call `toString` on `Promise.prototype.finally` but after importing Storybook (you have to use `require` in this instance, not `import`, so the ordering works for the test to see that `finally` starts off being an actual function) it’s no undefined.

Example repo here, https://github.com/benjaminreid/SBPromiseTest

**Expected behavior**
`Promise.prototype.finally` to no be removed.

**Screenshots**
N/A

**Code snippets**
`finally` is valid here, but after the inclusion of Storybook, it’s `undefined`.

```js
console.log('Before:', Promise.prototype.finally.toString());

const StoryBook = require('./storybook');

console.log('After:', Promise.prototype.finally);
```

**System:**
```
Environment Info:

  System:
    OS: macOS Mojave 10.14.6
    CPU: (4) x64 Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
  Binaries:
    Node: 10.16.0 - /var/folders/94/btf_scw12xbdnqv5dtv2t2hr0000gn/T/fnm-shell-8148567/bin/node
    Yarn: 1.17.3 - /usr/local/bin/yarn
    npm: 6.9.0 - /var/folders/94/btf_scw12xbdnqv5dtv2t2hr0000gn/T/fnm-shell-8148567/bin/npm
  Browsers:
    Chrome: 77.0.3865.90
    Firefox: 69.0.1
    Safari: 13.0.2
  npmPackages:
    @storybook/react-native: ^5.3.0-alpha.17 => 5.3.0-alpha.17
```

**Additional context**
In my instance `redux-persist` use this in their `onBeforeLift` callback.

### LINKED ISSUE #160: Storybook doesn't load when linkTo from '@storybook/addon-links' is imported
Storybook doesn't load when  importing  { linkTo } from '@storybook/addon-links'

**To Reproduce**
Setup fresh storybook project, 
with the default config. Couldn't load Storybook until this line on Welcome.stories.js was commented out

**Screenshots**
 
image

**System:**
Please paste the results of `npx -p @storybook/cli@next sb info` here:

Environment Info:

  System:
    OS: macOS 11.2.3
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
  Binaries:
    Node: 13.14.0 - ~/.nvm/versions/node/v13.14.0/bin/node
    Yarn: 1.19.1 - /usr/local/bin/yarn
    npm: 6.14.4 - ~/.nvm/versions/node/v13.14.0/bin/npm
  Browsers:
    Chrome: 90.0.4430.93
    Safari: 14.0.3
  npmPackages:
    @storybook/addon-actions: ^5.3 => 5.3.21
    @storybook/addon-knobs: ^5.3 => 5.3.21
    @storybook/addon-links: ^5.3 => 5.3.21
    @storybook/addon-ondevice-actions: ^5.3.23 => 5.3.23
    @storybook/addon-ondevice-knobs: ^5.3.25 => 5.3.25
    @storybook/react-native: ^5.3.25 => 5.3.25
    @storybook/react-native-server: ^5.3.23 => 5.3.23

### LINKED ISSUE #167: [6.0] update storybook dependencies to the latest stable versions.
in the next branch update the dependencies to the latest ones available.

### LINKED ISSUE #169: [6.0] update dependencies for ondevice addons to 6.0
the ondevice addons haven't had the storybook dependencies updated

- go to each addon in addons/ and update storybook dependencies such as @storybook/addons and @storybook/core-events

dannyhw:
Issue: #167 #169 #55 

should also fix #20 #160 #159

## What I did

Updated dependencies and metro config. The latest storybook alpha includes a modern build without polyfills that we can make use of by using the resolverMainfields option in metro config.

resolverMainFields: ['sbmodern', 'main']

With this change all issues with polyfills from storybook/addons and client should be resolved.
Going forward we need to replace knobs with controls because knobs don't support this and will break.

## How to test

In the example app I added a button under the button promise->finally story. If you press the "Do a promise" button and the app doesn't error then the problem is fixed.


addons/ondevice-actions/package.json
```diff
@@ -26,13 +26,13 @@
     "prepare": "node ../../scripts/prepare.js"
   },
   "dependencies": {
-    "@storybook/addons": "6.0.28",
```
oblador, line 29: Doesn't it make sense to be a bit more lenient with these versions for users (like me) that use storybook for multiple platforms? We can still lock it down to a certain/single version in the lockfile for this repo.

dannyhw: For example changing the addons dependencies to peer dependencies?

examples/native/.storybook/main.js
```diff
@@ -1,4 +1,3 @@
 module.exports = {
-  //stories: ['./components/*/*.stories.?(ts|tsx|js|jsx)', './globtest/Buttonx/Buttonx.stories.tsx'],
-  stories: ['./components/Button/Button.stories.tsx'],
+  stories: ['./components/*/*.stories.?(ts|tsx|js|jsx)', './globtest/Buttonx/Buttonx.stories.tsx'],
```
oblador: What's the purpose of `globtest/Buttonx/Buttonx.stories.tsx`?

addons/ondevice-actions/package.json
oblador: Not necessarily, but now you're essentially forcing consumers to use one single version of the `@storybook/*` packages, and it's not unlikely that they might have other dependencies wanting the same package and will thus end up with multiple versions and potentially incompatibilities.

examples/native/.storybook/main.js
dannyhw: This is something that I did previously, I wanted to test the story detection was working for multiple different cases. I think the folder name "globtest" is a bit misleading actually. Both these strings are just specifying locations of story files. Could easily just remove this example, however it's to show you can specify a full path.

addons/ondevice-actions/package.json
dannyhw: I don't have a particular reason to force a version. However I would say that with 6.0 there are going to be things that only work with 6.0+. 

I'm open to suggestions on how we can improve this experience for your use case. Perhaps you could make an issue that describes the use case and some potential solutions? Then we can discuss it further and include it in the 6.0 project board if it makes sense to do so.

examples/native/.storybook/main.js
dannyhw: fyi in this PR I just uncommented the line since I forgot to do so when I was testing the CSF example `['./components/Button/Button.stories.tsx'],` 😅.

addons/ondevice-actions/package.json
oblador: What do you think about setting it to `^6`?

oblador: I understand that the main storybook repo/package needs to keep their versions strict, but since this is maintained as an separate workspace/package with its own release cycle it shouldn't be strict IMHO.

dannyhw: Something like this?
```
  "dependencies": {
    "@storybook/addons": "^6",
    "@storybook/core-events": "^6",
    "core-js": "^3.0.1",
    "fast-deep-equal": "^2.0.1"
  },
  "devDependencies": {
    "@storybook/addon-actions": "^6"
  },
  "peerDependencies": {
    "@storybook/addon-actions": "*",
    "react": "*",
    "react-native": "*"
  },
```

honestly not sure why corejs is there tbh

dannyhw: Also in this PR I'm using the alpha version because thats where the new sb-modern field was introduced, so once 6.3 gets merged it would be possible to do ^6.

examples/native/.storybook/main.js
dannyhw: i chaged the name to be less strange :)

w0o:
Hey @dannyhw, 
this is an important fix myself an many will be thankful for and I wanna help test this. I tried (npm-) installing directly from git but as it doesn't include a dist/tarball I tried building it but hit some snags with Typescript errors across the example folders.

Any recommendation on how to test this code on a working project?
Any chance of seeing this released as part of a patch or new alpha/beta release for testing?

dannyhw:
@w0o this is already part of the v6 beta

Issue #186, opened 2021-06-25
Examples cannot be run on Android & macOS on the next-6.0 branch


lauriharpf:
**Describe the bug**
React Native Storybook examples cannot be run on Android & macOS with 3a199cce9cd8c4563683930247cc1777f02d4399 

**To Reproduce**
1. Clone the repository:

```
git clone https://github.com/storybookjs/react-native.git react-native-storybook
cd react-native-storybook
yarn bootstrap --core
yarn dev
```

2. In another terminal window:

```
cd examples/native
yarn
yarn android
```
3. Metro reports an error and the application fails to start:

```
Error: EISDIR: illegal operation on a directory, read
    at Object.readSync (fs.js:592:3)
    at tryReadSync (fs.js:366:20)
    at Object.readFileSync (fs.js:403:19)
    at UnableToResolveError.buildCodeFrameMessage (/Users/lauriharpf/Repos/lauriharpf/react-native-storybook/node_modules/metro/src/node-haste/DependencyGraph/ModuleResolution.js:347:17)
    at new UnableToResolveError (/Users/lauriharpf/Repos/lauriharpf/react-native-storybook/node_modules/metro/src/node-haste/DependencyGraph/ModuleResolution.js:333:35)
    at ModuleResolver.resolveDependency (/Users/lauriharpf/Repos/lauriharpf/react-native-storybook/node_modules/metro/src/node-haste/DependencyGraph/ModuleResolution.js:211:15)
    at DependencyGraph.resolveDependency (/Users/lauriharpf/Repos/lauriharpf/react-native-storybook/node_modules/metro/src/node-haste/DependencyGraph.js:413:43)
    at /Users/lauriharpf/Repos/lauriharpf/react-native-storybook/node_modules/metro/src/lib/transformHelpers.js:317:42
    at /Users/lauriharpf/Repos/lauriharpf/react-native-storybook/node_modules/metro/src/Server.js:1471:14
    at Generator.next (<anonymous>)
```

**Expected behavior**
The example app starts and examples can be used to verify functionality.

**Screenshots**
Screenshot of emulator with react-native 0.64.1 as the specified dependency in `/examples/native/package.json`:

![Screen Shot 2021-06-25 at 12 12 11](https://user-images.githubusercontent.com/6605505/123402619-97a40000-d5b0-11eb-8692-96ba32b8bc90.png)

Screenshot of emulator with react-native 0.64.2 as the specified dependency in `/examples/native/package.json` (see "Additional context"):

[...]

PR #188, opened 2021-06-25
Modified instructions for running example app in the repo

lauriharpf:
Issue: https://github.com/storybookjs/react-native/issues/186

## What I did
Modified instructions to run example apps to match discussion in https://github.com/storybookjs/react-native/issues/186 and current next-6.0 branch content

## How to test
Follow the instructions and confirm that the example works as expected.

## Additional details
Also removed the parts referring to the Expo app, which doesn't exist in `next-6.0` at the moment. It exists in the `master` branch, so didn't remove the mention from the "Reproductions" section. If the `crna-kitchen-sink` example is later ported to 6.0, then that part of the documentation should be re-added.


dannyhw (approved):
Looks good to me. Thanks for your contribution 🙂 .

Issue #177, opened 2021-05-31
[6.0] ondevice-controls


dannyhw:
Now that args and CSF are supported its time to add in controls support.

Controls will 
- allow for changing props in real time similar to knobs
- detect prop types automatically
- utilise args to provide a simpler api than knobs

PR #197, opened 2021-07-02
feat: First initial poking towards re-adding Select control

lauriharpf:
Issue: #177

## What I did
This is just a first, **very** initial and rudimentary draft towards re-enabling `Select`. **Note that it doesn't work** at its current state, just sharing it as a "hey, this is where I left off"-state in case there would be something that would help. Don't seem to get the options as props and couldn't yet figure out why.

While working on this, I noticed that controls seems to now have an "Options" structure in the main storybook (see https://github.com/storybookjs/storybook/blob/6584de334f4159305adc23ef73ac9e62e59833c2/lib/components/src/controls/options/Options.tsx). So this probably needs a similar structure?

(Note also the xxxExample-naming; this was done prior to our discussion in another PR; lets go with what we decided there)

## How to test

- Does this need a new example in examples/native? **Yes, we need to add one**
- Does this need an update to the documentation? **Not sure**


dannyhw:
just pushed a fix for the select control.

PR #218, opened 2021-07-13
fix: Fixes #211 (After opening the keyboard the android UI shrinks)

### LINKED ISSUE #211: [6.0] After opening the keyboard the android UI shrinks
This behaves in a strange way for me on the Android Emulator, can you reproduce it? Seems to work fine on the iOS emulator, can't reproduce the issue there.

1. Open the story Navigator.
2. Select the "Basic" story.
3. Open "Addons -> Controls"
4. Click on the "first" field so that the keyboard comes up.
5. Click back to "Navigator"

## What happens
The story list in the navigator gets truncated a lot (seems to only occupy ~50% or so of the screen) after Step 5.

## Expected result
The story list would be rendered in the same way in Step 5 as it was in Step 1.

### Screenshot after Step 1
Screen Shot 2021-07-11 at 20 08 59

### Screenshot after Step 4
Screen Shot 2021-07-11 at 20 11 45

### Screenshot after Step 5
Screen Shot 2021-07-11 at 20 13 15

_Originally posted by @lauriharpf in https://github.com/storybookjs/react-native/pull/182#discussion_r667511053_

lauriharpf:
Issue: #211 

## What I did
Removed 

## How to test
The steps to reproduce this issue on the Android emulator were:

0. Run the React Native examples (under examples/native) on Android
1. Open the story Navigator.
2. Select the "Basic" story.
3. Open "Addons -> Controls"
4. Click on the "first" field so that the keyboard comes up.
5. Click back to "Navigator"

### What happened
The story list in the navigator gets truncated a lot (seems to only occupy ~50% or so of the screen) after Step 5.

### What happens now after the fix
The story list doesn't get truncated; it looks the same it did after Step 1.

### Other

- Does this need a new example in examples/native? **no**
- Does this need an update to the documentation? **no**

The code this PR removes was introduced in https://github.com/storybookjs/react-native/commit/e413ec2e05103033e0db7e1e33f7c0a4e988fae6 . While testing (both on Android and iOS), I didn't notice any issues caused by removing this code. It is of course possible that there is a case that I've missed, but the code this PR removes clearly causes an issue in a common use case.

### UX with this branch
https://user-images.githubusercontent.com/6605505/125475611-ab4deae6-3873-4ed6-95f4-974c2a356c96.mov

### UX in the next-6.0 branch (=without this fix)
https://user-images.githubusercontent.com/6605505/125475678-2a4ccee1-00e6-460d-8c92-b42d4dab2280.mov

### DIFF
--- a/examples/native/ios/Podfile.lock
+++ b/examples/native/ios/Podfile.lock
@@ -459,7 +459,7 @@ SPEC CHECKSUMS:
   CocoaAsyncSocket: 065fd1e645c7abab64f7a6a2007a48038fdc6a99
   DoubleConversion: cf9b38bf0b2d048436d9a82ad2abe1404f11e7de
   FBLazyVector: 7b423f9e248eae65987838148c36eec1dbfe0b53
-  FBReactNativeSpec: 60baaee9d10a9d225389062decbbf2b0bd6420ce
+  FBReactNativeSpec: ba3bc03e12cb0bea22d69a8a9458eaf3e92521a8
   Flipper: d3da1aa199aad94455ae725e9f3aa43f3ec17021
   Flipper-DoubleConversion: 38631e41ef4f9b12861c67d17cb5518d06badc41
   Flipper-Folly: 755929a4f851b2fb2c347d533a23f191b008554c
--- a/app/react-native/src/preview/components/OnDeviceUI/absolute-positioned-keyboard-aware-view.tsx
+++ b/app/react-native/src/preview/components/OnDeviceUI/absolute-positioned-keyboard-aware-view.tsx
@@ -1,13 +1,5 @@
-import React, { ReactNode, useRef, useEffect } from 'react';
-import {
-  Platform,
-  Keyboard,
-  Dimensions,
-  View,
-  LayoutChangeEvent,
-  KeyboardEvent,
-  StyleSheet,
-} from 'react-native';
+import React, { ReactNode } from 'react';
+import { View, LayoutChangeEvent, StyleSheet } from 'react-native';
 
 export interface PreviewDimens {
   width: number;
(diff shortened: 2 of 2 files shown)


dannyhw:
How does this behave when using an input in the preview (as part of a story)?

Thanks again for your contribution :)

examples/native/ios/Podfile.lock
```diff
@@ -459,7 +459,7 @@ SPEC CHECKSUMS:
   CocoaAsyncSocket: 065fd1e645c7abab64f7a6a2007a48038fdc6a99
   DoubleConversion: cf9b38bf0b2d048436d9a82ad2abe1404f11e7de
   FBLazyVector: 7b423f9e248eae65987838148c36eec1dbfe0b53
-  FBReactNativeSpec: 60baaee9d10a9d225389062decbbf2b0bd6420ce
+  FBReactNativeSpec: ba3bc03e12cb0bea22d69a8a9458eaf3e92521a8
```
dannyhw, line 462: there is a bug with react native 64.1 where this hash is different on every machine, it's probably better not to commit these changes for now (FBReactNativeSpec). Maybe we should add the podfile lock to the ignore?

dannyhw (approved):
I tried it out and its working fine for me. Great work 👍

examples/native/ios/Podfile.lock
dannyhw: Since this isn't really a blocker I merged anyway.

lauriharpf:
> How does this behave when using an input in the preview (as part of a story)?

Duh, you mentioned that in the issue description but I totally forgot. Sorry!

Glad to hear it worked for you. I could still make another PR that contains a story with a `TextInput` or two - those could be useful in confirming that functionality such as this works as intended? 🤔

dannyhw:
> > How does this behave when using an input in the preview (as part of a story)?
> 
> Duh, you mentioned that in the issue description but I totally forgot. Sorry!
> 
> Glad to hear it worked for you. I could still make another PR that contains a story with a `TextInput` or two - those could be useful in confirming that functionality such as this works as intended? 🤔

@lauriharpf I tried it out and it seems fine to me. I could commit the input example I already made though

lauriharpf:
> @lauriharpf I tried it out and it seems fine to me. I could commit the input example I already made though

Sounds great, please do! :+1: 🥇

examples/native/ios/Podfile.lock
lauriharpf: Thanks for the heads-up, wasn't aware of this! :+1:

I've been under the impression that version controlling Podfile.lock is a good idea (https://stackoverflow.com/questions/44028199/why-should-you-include-podfile-lock-under-version-control), but not sure how important it is in a React Native project 🤔 .

In any case, definitely no reason to commit the `FBReactNativeSpec` hash changes if it always changes from machine to machine at the moment - will keep that in mind!

Issue #41, opened 2020-04-10
keyboard auto dismiss


jb9229:
**Describe the bug**
1. When I try to write on the Text Input Commonent, the keyboard goes up and down quickly.

**Screenshots**
![스크린샷 2020-04-10 오후 3 17 22](https://user-images.githubusercontent.com/3200647/78968446-93243080-7b3f-11ea-89d5-d2c49174b04a.png)

**Additional context**
1. Android Env is okay
2. Production Env is okay ios and android
3. This only happens in ios environments(actually i did't check ios real device )
4. React-native version: 0.61.2
4. @storybook version: v5.3.17

is there any solution?

Gandhi89:
facing same error

FrederickEngelhardt:
@dannyhw 
Same issue. Any solutions discovered?

My hope was to use the device keyboard to demo some features. In order to continue development I had to disable the device keyboard and use hardware keyboard.

Pressing CMD+K will cause the same behavior as noted above.

<img width="349" alt="Screen Shot 2021-06-09 at 9 46 08 AM" src="https://user-images.githubusercontent.com/31741411/121395709-8e0b6e80-c907-11eb-8f2f-e3cf210d5896.png">

dannyhw:
I will take a look tomorrow and see what I can find.

lauriharpf:
Did some testing, but couldn't reproduce this 🤔 .

## Next-6.0 branch (commit 69c176a5f3af3b52bfcc12745c9daad77432d4ba)
Couldn't reproduce the issue here with controls. At approx. 32 seconds pressed CMD+K to disable the device keyboard for a while, then re-enabled it with CMD+K at approx. 37 seconds.

https://user-images.githubusercontent.com/6605505/125478858-1f715cc4-c1aa-42dd-bf77-d358c7910a81.mp4

## master branch (commit b508563001597c0a2fd449437241d1d56e64b090)
Couldn't reproduce the issue here with knobs. Tested it on the Expo examples (`crna-kitchen-sink`) as I couldn't get the `native` examples to run on iOS even after some tweaking. At approx. 28 seconds pressed CMD+K to disable the device keyboard for a while, then re-enabled it with CMD+K at approx. 38 seconds.

https://user-images.githubusercontent.com/6605505/125486219-5574dadb-364d-42fd-ad82-1001432bdd08.mp4

Also attempted to reproduce this by adding a `TextInput` to a story, but that seemed to work too.

[...]

PR #220, opened 2021-07-13
feat: add text input story to example

dannyhw:
Issue:

## What I did

I've added a new text-input example that can be used to test for issues with the device keyboard.

this should help with testing issues like #41


dannyhw:
@lauriharpf do you think I should add more to this example or does it cover everything?

lauriharpf (approved):
This looks good & sufficient at this stage to me :+1: 🍪  🍰  🏅 . Good work! Let's expand later when/if we see a need to do so.

Issue #199, opened 2021-07-11
[6.0] Add remaining control types from knobs to ondevice controls


dannyhw:
In #182 the core of ondevice controls is added. 

This piece of work is to re add the remaining controls.

The remaining controls to re-add are

ArrayType
DateType
ButtonType
RadioType

dannyhw:
Working on ArrayType

dannyhw:
working on the DateType

dannyhw:
working on button/radio

PR #221, opened 2021-07-13
feat: add array control type

dannyhw:
Issue: #199 

## What I did

I added the Array control type and an associated example

## How to test

Run the example and open the array story

### DIFF
--- a/examples/native/components/ControlExamples/Array/Array.stories.tsx
+++ b/examples/native/components/ControlExamples/Array/Array.stories.tsx
@@ -1,0 +1,27 @@
+import React from 'react';
+import { ComponentStory, ComponentMeta } from '@storybook/react-native';
+import { Array } from './Array';
+
+const ArrayMeta: ComponentMeta<typeof Array> = {
+  title: 'Array',
+  component: Array,
+  args: {
+    list: ['a', 'b', 'c'],
+  },
+  argTypes: {
+    list: {
+      separator: ',',
+      control: { type: 'array' },
+    },
+  },
+  parameters: {
+    notes:
+      'Seems like the array type is not distinguishable from object so you should provide the arg type in this case',
+  },
+};
+
+export default ArrayMeta;
+
+type ArrayStory = ComponentStory<typeof Array>;
+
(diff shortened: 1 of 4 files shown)


dannyhw:
@lauriharpf if you have time could you review these changes?

examples/native/components/ControlExamples/Array/Array.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { Text, View, StyleSheet } from 'react-native';
+
+interface ArrayProps {
+  list: string[];
+}
+
+export const Array = ({ list }: ArrayProps) => (
+  <View style={styles.container}>
+    {list.map((item) => (
```
lauriharpf: Could we do something like 

```typescript
    {list.map((item, index) => (
      <Text key={index} style={styles.item}>
        {item}
      </Text>
    ))}
```

here to avoid the warning?

Screen Shot 2021-07-14 at 21 34 04

examples/native/components/ControlExamples/Array/Array.stories.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { ComponentStory, ComponentMeta } from '@storybook/react-native';
+import { Array } from './Array';
+
+const ArrayMeta: ComponentMeta<typeof Array> = {
+  title: 'Array',
```
lauriharpf: The other controls seem to have "{name} control" titles, so it could be a bit tidier if this was uniform with them ("Array control").

lauriharpf (approved):
Looks nice & works well on both platforms :+1: 🚀 🥇 !

Just a couple of very minor comments, nothing that blocks this IMO.

examples/native/components/ControlExamples/Array/Array.tsx
dannyhw: Yes! I forgot about this, thanks :)

examples/native/components/ControlExamples/Array/Array.stories.tsx
dannyhw: Makes sense, I'll update this.

PR #222, opened 2021-07-13
fix: Fixes #219 (editing a story causes a full reload of the app)

### LINKED ISSUE #219: [6.0] When editing a story the app will always fully reload
**Describe the bug**

When editing a story fast reload is not working correctly.

This section of code was removed from the loadCsf file and might be related (line 251)

```
 if (m && m.hot && m.hot.dispose) {
      ({ previousExports = new Map() } = m.hot.data || {});
      m.hot.dispose((data) => {
        loaded = false;
        // eslint-disable-next-line no-param-reassign
        data.previousExports = previousExports;
      });
      m.hot.accept();
    }
```

The dispose part is not need however perhaps maintaining the hot.accept part would be useful... this needs some investigation probably.
```
 if (m && m.hot) {
      m.hot.accept();
  }
 
```

**To Reproduce**

1. run example app
2. edit a story file and save it or just save with no changes
2. app fully reload

**Expected behavior**

Fast reload should be used such that only what changed gets reloaded.

lauriharpf:
# Note: Breaks fast reloading of storiesOf -format
With this change, editing the `storiesOf`-format stories (e.g. `Button.stories.tsx` under `/examples/native/components/PromiseTest`) starts throwing an error (see screenshot).

Don't know why that happens. Perhaps fast reloading `Button.stories.tsx` causes the `storiesOf(..).add` call to get re-run? Adding a `storyStore.startConfiguring()` call (see https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#story-store-immutable-outside-of-configuration ) prior to `m.hot.accept()` does not help.

Screen Shot 2021-07-14 at 11 47 18

## UX for CSF stories the next-6.0 branch (=without this fix)

https://user-images.githubusercontent.com/6605505/125520359-da2fe881-c503-47b5-b78b-04152ad990aa.mov

## UX for CSF stories this branch

https://user-images.githubusercontent.com/6605505/125580674-10993194-3246-456e-b3e6-557a78ebdc91.mov

Issue: #219 

## What I did
Changed how the CSF stories get reloaded.

Had to use the `FORCE_RE_RENDER` workaround to get fast reload changes to be immediately reflected on the UI.

## How to test
See the steps to reproduce the issue at https://github.com/storybookjs/react-native/issues/219. Basically, edit a CSF story (e.g. `Boolean.stories.tsx`) while the app is running.

### DIFF
--- a/app/react-native/src/preview/loadCsf.ts
+++ b/app/react-native/src/preview/loadCsf.ts
@@ -244,5 +244,10 @@ export const loadCsf = ({
       );
     }
 
+    if (m && m.hot) {
+      storyStore.clearGlobalDecorators();
+      m.hot.accept();
+    }
+
     configApi.configure(loadStories(loadable, framework, { clientApi, storyStore }), m);
   };
--- a/app/react-native/src/preview/loadCsf.ts
+++ b/app/react-native/src/preview/loadCsf.ts
@@ -1,5 +1,6 @@
 import { ClientApi, ConfigApi, StoryStore } from '@storybook/client-api';
 import { logger } from '@storybook/client-logger';
+import { FORCE_RE_RENDER } from '@storybook/core-events';
 import { isExportStory, storyNameFromExport, toId } from '@storybook/csf';
 import './global';
 
@@ -247,6 +248,7 @@ export const loadCsf = ({
     if (m && m.hot) {
       storyStore.clearGlobalDecorators();
       m.hot.accept();
+      storyStore._channel.emit(FORCE_RE_RENDER);
     }
 
     configApi.configure(loadStories(loadable, framework, { clientApi, storyStore }), m);


intergalacticspacehighway:
@lauriharpf I tried your changes. It works in version 6! Thank you. I faced one issue while testing in version 6. It clears all the decorators in the initial load too. Do you have any pointers on why that might be happening?

dannyhw:
@intergalacticspacehighway can you explain what you mean by it clearing all the decorators? What do you see happening?

intergalacticspacehighway:
@dannyhw thanks for quick response. I think [this line](https://github.com/storybookjs/react-native/pull/222/files#diff-56f7bae1eb7c55e992b2633596dfddde2b4f1352dd1802cdc60c675b8d047e95R249) clears the decorators.

I am using a ThemeProvider as a decorator as shown below.
```jsx
export const decorators = [
	Story => (
		<DripsyProvider theme={theme}>
			<Story />
		</DripsyProvider>
	),
]
```

When I tried this PR, the stories where not able to access the Provider. i.e. the decorator was getting removed. Interestingly when I tried removing [this line](https://github.com/storybookjs/react-native/pull/222/files#diff-56f7bae1eb7c55e992b2633596dfddde2b4f1352dd1802cdc60c675b8d047e95R249), stories were able to access the Provider, but on refresh I get a warning from [this line](https://github.com/storybookjs/storybook/blob/main/lib/client-api/src/story_store.ts#L314) that said you cannot add a duplicate decorator.

I hope this answers your question. Let me know if you need a reproducible demo. Thank you!

dannyhw:
fixed in #384

PR #224, opened 2021-07-14
fix: Fixes selected story changing when editing the story file

lauriharpf:
Issue: #219 

## What I did
Fixed issue that could be repeated with the following steps:

1. Run the React Native examples (under examples/native)
2. Select a story, e.g. "Boolean control - Basic" (Boolean.stories.tsx)
3. Edit that story file (e.g. change the "on" argument to false in Boolean.stories.tsx)

### What happened before this fix
The entire app reloaded and the currently selected story was changed. See video below:

https://user-images.githubusercontent.com/6605505/125652354-07599d01-9c15-42de-9c5d-992193705e2a.mov

### What happens now after the fix
The entire app reloads, but the selected story persists. This improves the developer UX (don't need to always re-select the story after changes).

https://user-images.githubusercontent.com/6605505/125652509-dbba06ab-b599-4c2f-b2ac-55bebaab4065.mov

Unlike PR #222 this also works with stories that are in `storiesOf` format:

https://user-images.githubusercontent.com/6605505/125652694-fb0b313c-8407-4c6d-898e-b143fcfe2aec.mov

## How to test
See example steps under "What I did".

- Does this need an update to the documentation? **no*', not in my opinion

### DIFF
--- a/app/react-native/src/preview/Preview.tsx
+++ b/app/react-native/src/preview/Preview.tsx
@@ -84,11 +84,12 @@ export default class Preview {
       );
     }
 
-    const { initialSelection, shouldPersistSelection } = params;
-    this._setInitialStory(initialSelection, shouldPersistSelection);
     if (params.asyncStorage) {
       this._asyncStorage = params.asyncStorage;
     }
+
+    const { initialSelection, shouldPersistSelection } = params;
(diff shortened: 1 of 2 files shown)


dannyhw (approved):
Seems to work well, definitely improves a lot on the current behaviour.

Thanks again 🙏

PR #225, opened 2021-07-14
refactor: Potential refactor of the Date control

lauriharpf:
Just a thought, see how you feel.

Two different boolean values ("isDateVisible" and "isTimeVisible") allow for one invalid state (both shouldn't be true at the same time, as that would mean both pickers are visible simultaneously, which can't happen with the current implementation).


dannyhw (approved):
Looks good to me, thanks for this 👍

Issue #213, opened 2021-07-12
Rework scripts in the mono repo and update tooling


dannyhw:
We need to take a look at the scripting for various things in the repository to make sure it all works considering the changes that have been made for 6.0.

These changes are things like moving to typescript for all packages and some of the changes that were included in #182. In that PR some changes were made to the way ondevice controls and @storybook/react-native are compiled and those packages should be consistent with the others.

## Goals

- Verify that controls and storybook/react-native will work with the release process after #182
- investigate existing scripts to identify improvements and either raise an issue or implement those changes

In #182 there was some discussion that should be considered when making changes
- should only copy what is necessary - [https://github.com/storybookjs/react-native/pull/182#discussion_r667367144](https://github.com/storybookjs/react-native/pull/182#discussion_r667367144)
- should be platform independent - [https://github.com/storybookjs/react-native/pull/182#discussion_r667365995](https://github.com/storybookjs/react-native/pull/182#discussion_r667365995)

PR #226, opened 2021-07-14
feat: adds a prepareall script

dannyhw:
Issue: #213 

## What I did

I added a script that you can use to run the prepare command on all packages. This has been useful for when i don't want to run the watcher.

The way I use this is that when I want to test a PR I check out the PR and run 

```
yarn install
yarn prepareAll
cd examples/native/ios
pod install
```

This way everything is ready to test and I just need to run the app.

Open to feedback and ideas!

## How to test

run `yarn prepareAll`


lauriharpf (approved):
Checked this out, seems to work really nicely. A bit curious though: Why don't you want to always run the watcher? Are there some downsides to it?

dannyhw:
> Checked this out, seems to work really nicely. A bit curious though: Why don't you want to always run the watcher? Are there some downsides to It?

Well it just depends what I'm doing. If I'm working on a feature I'll sometimes use the watcher because it's convenient.

However if I'm testing a pr I usually don't plan on changing anything and I just want to make sure everything is ready to test. I could run the watcher but I prefer to keep down the number of background processes when possible.

Whenever I run the Android emulator my Mac is already struggling 😅

PR #228, opened 2021-07-15
fix: Fixes the "`storyFn` is deprecated and will be removed in Storybook 7.0." deprecation warning 

lauriharpf:
Fixes the storyFn deprecation warning that was shown when rendering the story. Fix based on the instructions at https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-storyfn

Screen Shot 2021-07-15 at 16 10 57

Issue: (Don't know if there is one; could be 6.0-specific?)

## What I did
Changed `StoryView.tsx` with instructions from https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-storyfn .

## How to test
Experiment with the different example stories in `examples/native`: They should work as previously, just the deprecation warning should be gone.

### DIFF
--- a/app/react-native/src/preview/components/StoryView/StoryView.tsx
+++ b/app/react-native/src/preview/components/StoryView/StoryView.tsx
@@ -1,10 +1,11 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 
 import { StoreItem } from '@storybook/client-api';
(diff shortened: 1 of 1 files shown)


dannyhw (approved):
Nice one! Works perfectly :)

Issue #200, opened 2021-07-11
[6.0] reintroduce all ondevice addons into example and verify functionality with 6.0


dannyhw:
In #182 some of this work was completed however backgrounds and actions havent been looked at yet. 

We should re-add all addons to the example and include example usage.

PR #229, opened 2021-07-16
feat: Reintroduce the backgrounds addon 

lauriharpf:
Issue: #200 (part of the scope of that issue)

## What I did
Reintroduced the backgrounds addon as a part of completing #200. Includes both storiesOf and CSF examples of usage.

## How to test
Run the React Native examples under `examples/native` on the iOS and Android emulators. Experiment with the added examples:

* **Background StoriesOf**: Not sure if we want to keep this, given that **storiesOf** is apparently going away? However, it can be used to demo that this works with existing stories that use the **storiesOf** format.
* **Background CSF**: Same example in the CSF format.

Also check the modified "Backgrounds not configured" text (e.g. open button promise > finally > Addons > Backgrounds).

- Does this need a new example in examples/native? **yes**, added both CSF & StoriesOf examples
- Does this need an update to the documentation? **yes**; did a minor update to the `@storybook/addon-ondevice-backgrounds` README, but we should have much better docs & also confirm/add support for all of the web examples

### Screen recording of functionality

https://user-images.githubusercontent.com/6605505/125938162-3b33027e-5067-41cc-8e18-63b66ec461f1.mp4

### ~~Known issue: Reloading on iOS does not work~~
Edit: Never mind, did some changes in 73e783a2f3467f830c74f7b508e31f44952c93dd to fix this.
~~Reloading the app (e.g. via the emulator or due to making changes) causes the controls in "Addons => Backgrounds" to disappear on iOS. It works correctly on Android. Don't know what is causing this. Any help in fixing is appreciated. See recording below:~~

https://user-images.githubusercontent.com/6605505/125938242-6f94e413-1d9d-4938-b006-ac7f220507e6.mov

### DIFF
--- a/addons/ondevice-backgrounds/README.md
+++ b/addons/ondevice-backgrounds/README.md
@@ -2,8 +2,6 @@
 
 Storybook Backgrounds Addon for react-native can be used to change background colors of your stories right from the device.
 
-<img src="https://raw.githubusercontent.com/storybookjs/storybook/master/addons/ondevice-backgrounds/docs/demo.gif" alt="Storybook Backgrounds Addon Demo" width="400" />
-
 ## Installation
 
 ```sh
@@ -12,41 +10,16 @@ yarn add -D @storybook/addon-ondevice-backgrounds
 
 ## Configuration
 
-Create a file called `rn-addons.js` in your storybook config.
-
-Add following content to it:
+Then, add following content to `.storybook/main.js`:
 
 ```js
-import '@storybook/addon-ondevice-backgrounds/register';
-```
-
-Then import `rn-addons.js` next to your `getStorybookUI` call.
-
-```js
-import './rn-addons';
+module.exports = {
+  addons: ['@storybook/addon-ondevice-backgrounds'],
+};
 ```
 
 ## Usage
 
-react-native users will have to import `storiesOf` from `@storybook/react-native` and are required to add the `withBackgrounds` decorator.
+See the [example of using the Backgrounds addon with Component Story Format](../../examples/native/components/BackgroundExample/BackgroundCsf.stories.tsx). You can also run the [react-native app](../../examples/native) to see it in action.
 
-Then write your stories like this:
-
-```js
-import React from 'react';
-import { storiesOf } from '@storybook/react-native';
-import { withBackgrounds } from '@storybook/addon-ondevice-backgrounds';
-
-addDecorator(withBackgrounds);
-
-storiesOf('Button', module)
-  .addParameters({
-    backgrounds: [
-      { name: 'dark', value: '#222222' },
-      { name: 'light', value: '#eeeeee', default: true },
-    ],
-  })
-  .add('with text', () => <Text>Click me</Text>);
-```
-
-See [web backgrounds addon](../backgrounds#usage) for detailed usage and the [crna-kitchen-sink app](../../examples/crna-kitchen-sink) for more examples.
+The [web backgrounds addon documentation](https://storybook.js.org/docs/react/essentials/backgrounds) may also be useful, but the examples there have not been tested with React-Native Storybook.
--- a/examples/native/.storybook/main.js
+++ b/examples/native/.storybook/main.js
@@ -3,5 +3,9 @@ module.exports = {
     './components/**/*.stories.?(ts|tsx|js|jsx)',
     './other_components/AnotherButton/AnotherButton.stories.tsx',
   ],
-  addons: ['@storybook/addon-ondevice-notes', '@storybook/addon-ondevice-controls'],
+  addons: [
+    '@storybook/addon-ondevice-notes',
+    '@storybook/addon-ondevice-controls',
+    '@storybook/addon-ondevice-backgrounds',
+  ],
 };
--- a/examples/native/components/BackgroundExample/BackgroundCsf.stories.tsx
+++ b/examples/native/components/BackgroundExample/BackgroundCsf.stories.tsx
@@ -1,16 +1,13 @@
 import React from 'react';
 import { ComponentStory, ComponentMeta } from '@storybook/react-native';
 import { withBackgrounds } from '@storybook/addon-ondevice-backgrounds';
-import { Dimensions, Text, View, StyleSheet } from 'react-native';
+import { Text, StyleSheet } from 'react-native';
 
 const Background = () => (
-  <View style={styles.view}>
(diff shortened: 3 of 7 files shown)


examples/native/components/BackgroundExample/BackgroundCsf.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { Dimensions, Text, View, StyleSheet } from 'react-native';
+
+const Background = () => (
+  <View style={styles.view}>
+    <Text style={styles.text}>Change background color via Addons -&gt; Background</Text>
+  </View>
+);
+
+const styles = StyleSheet.create({
+  view: { height: Dimensions.get('window').height },
```
lauriharpf: I don't know why this is necessary: For some reason, the text refused to appear until I added this View with an explicit definition of the height. This hack is also included in the example in `BackgroundPanel.tsx` (which is a copy-paste of this file).

Didn't see the same phenomenon with the other CSF stories. Also the `storiesOf` example just renders the `<Text>` component without any wrappers and it seems to work fine 🤔

examples/native/.storybook/main.js
```diff
@@ -3,5 +3,9 @@ module.exports = {
     './components/**/*.stories.?(ts|tsx|js|jsx)',
     './other_components/AnotherButton/AnotherButton.stories.tsx',
   ],
-  addons: ['@storybook/addon-ondevice-notes', '@storybook/addon-ondevice-controls'],
+  addons: [
+    '@storybook/addon-ondevice-notes',
+    '@storybook/addon-ondevice-controls',
+    '@storybook/addon-ondevice-backgrounds',
```
lauriharpf, line 9: It's interesting that this seems to work even though there's no explicit dependency at https://github.com/lauriharpf/react-native/blob/feat/reintroduce-backgrounds-addon/examples/native/package.json

`yarn why` says this:

```
lauriharpf@Lauris-MacBook-Pro native % yarn why @storybook/addon-ondevice-backgrounds
yarn why v1.22.10
[1/4] 🤔  Why do we have the module "@storybook/addon-ondevice-backgrounds"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "@storybook/addon-ondevice-backgrounds@6.0.0-alpha.0"
info Has been hoisted to "@storybook/addon-ondevice-backgrounds"
info Reasons this module exists
   - "workspace-aggregator-308ac265-8f7c-4c55-8d6d-d9a9373c67cf" depends on it
   - Specified in "dependencies"
   - Hoisted from "_project_#@storybook#addon-ondevice-backgrounds"
info Disk size without dependencies: "1.66MB"
info Disk size with unique dependencies: "10.8MB"
info Disk size with transitive dependencies: "34.95MB"
info Number of shared dependencies: 81
```

Should we still explicitly add the dependency to the `examples/native` project (`yarn add -D @storybook/addon-ondevice-backgrounds`) so that people don't get confused?

dannyhw: Yeah it should be in the package.json in examples/native. Probably its only working because of the metro config that adds the packages to the watcher.

addons/ondevice-backgrounds/README.md
```diff
@@ -12,41 +10,16 @@ yarn add -D @storybook/addon-ondevice-backgrounds
...
-```js
-import '@storybook/addon-ondevice-backgrounds/register';
-```
-
-Then import `rn-addons.js` next to your `getStorybookUI` call.
+Then, add following content to `.storybook/main.js`:
 
 ```js
-import './rn-addons';
+module.exports = {
```
dannyhw, line 16: Nice job on updating the documentation! This is something I forgot to do on controls and notes 😅 .

dannyhw:
Can you add an example of the parameter/decorator to the preview.js file? That would apply a global decorator and parameter.

Not entirely sure how necessary the decorator is but it would be something like this 

examples/native/.storybook/preview.js
```
import React from 'react';
import { View, StyleSheet } from 'react-native';
import { withBackgrounds } from '@storybook/addon-ondevice-backgrounds';

export const decorators = [
  (StoryFn) => (
    <View style={styles.container}>
      <StoryFn />
    </View>
  ),
  withBackgrounds,
];
export const parameters = {
  my_param: 'anything',
  backgrounds: [
    { name: 'warm', value: 'hotpink', default: true },
    { name: 'cool', value: 'deepskyblue' },
  ],
};

const styles = StyleSheet.create({
  container: { padding: 8 },
});

```

examples/native/.storybook/main.js
lauriharpf: Added the dependency to the examples/native package.json :+1:

lauriharpf:
> Can you add an example of the parameter/decorator to the preview.js file? That would apply a global decorator and parameter.

Thanks, did that 👍 . The only downside of having it globally is that the "backgrounds not configured" functionality (shown in the screen recording at ~0:30) is not visible in `examples/native` anymore.

dannyhw:
> > Can you add an example of the parameter/decorator to the preview.js file? That would apply a global decorator and parameter.
> 
> Thanks, did that 👍 . The only downside of having it globally is that the "backgrounds not configured" functionality (shown in the screen recording at ~0:30) is not visible in `examples/native` anymore.

@lauriharpf ah ok, mostly I just wanted to make sure that decorators/params functionality was working. I think there might be a way to make that message come up though. Perhaps if you replace the backgrounds param with an empty list on one of the stories, not sure though. Either way I think the normal use case is to have this set globally so it's nice to have this example.

examples/native/components/BackgroundExample/BackgroundCsf.stories.tsx
dannyhw: So I found out what the problem is, the decorator in preview.js with the padding has no height. If you give it the flex:1 style then the text here will work as normal.

dannyhw: I committed the fix to your branch :)

dannyhw (approved):
It's working great, impressive work! 🎉

examples/native/components/BackgroundExample/BackgroundCsf.stories.tsx
lauriharpf: Awesome, thanks for fixing it and for explaining what was happening! 👍  🙏

PR #231, opened 2021-07-16
feat: radio control

### LINKED ISSUE #199: [6.0] Add remaining control types from knobs to ondevice controls

dannyhw:
Issue: closes #199 

## What I did

I re-added the radio control.

I was going to add the button as well but it seems pretty useless, I can't think of an example where it is useful. So I just removed it instead.

image

## How to test

1. Run the example app 
2. open the radio example
3. update the text by using the radio control in the addons panel.


examples/native/components/ControlExamples/Radio/Radio.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+const RadioMeta: ComponentMeta<typeof Radio> = {
+  title: 'Radio control',
+  component: Radio,
+  args: {
+    selection: radio_stations[0],
+  },
+  argTypes: {
+    selection: {
+      options: radio_stations,
+      control: { type: 'radios' },
```
lauriharpf, line 16: Wouldn't do anything in the scope of this PR, but longer-term we could consider aligning the control types with the control types of the Web storybook: https://storybook.js.org/docs/react/essentials/controls#annotation

They seem to have a bit different setup with the controls, especially with the `options` ones. Being uniform could help via e.g. allowing us to point to the same docs and making the library less confusing to use. This would be a longer-term V6 target though, not an Alpha 1 target.

Screen Shot 2021-07-17 at 10 33 45

examples/native/components/ControlExamples/Radio/Radio.stories.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { ComponentStory, ComponentMeta } from '@storybook/react-native';
+import { Radio } from './Radio';
+
+const radio_stations = ['104.8MHz', '909 kHz', '90FM'];
```
lauriharpf, line 5: I see what you did here 😂 👍

lauriharpf (approved):
Confirmed that this works well on both the iOS and Android platforms. Code looks clean, great job all in all! 👍 🥇 🚀

examples/native/components/ControlExamples/Radio/Radio.stories.tsx
dannyhw: Yeah I totally agree with you on this, mostly at this stage I just wanted to release something that would replace knobs and get the fundamentals of controls in. 

I made an issue for all the remaining controls work and part of that is basically  getting it closer to the web implementation.

Issue #237, opened 2021-07-18
[6.0] util and util-deprecate are not listed as dependencies and break unless installed


dannyhw:
**Describe the bug**

When setting up with the 6.0 alpha util and util-deprecate must be installed or the app breaks. Probably they need to be included in the @storybook/react-native package.json dependencies.

**To Reproduce**

Follow the steps at https://github.com/storybookjs/react-native/blob/next-6.0/v6README.md without installing util and util deprecate and you should get an error.

**Expected behavior**

There should be no error when not explicitly installing util/util-deprecate. These should be installed when @storybook/react-native is installed.

PR #245, opened 2021-07-27
docs: Remove the need to install util and util-deprecate from the v6README (works without them)

lauriharpf:
Issue: #237 

## What I did.
Confirmed that `util` and `util-deprecate` no longer need to be added as dependencies with alpha.1, so removed those from the v6README. **Note**: https://gist.github.com/dannyhw/9b84973dcc6ff4fa2e86e32d571d294e should also be edited accordingly.

## How to test
Repeat the steps in the v6README and confirm that the project works.

### DIFF
--- a/v6README.md
+++ b/v6README.md
@@ -21,7 +21,6 @@ Open up your react native project in your chosen editor, here I use vscode
 code .
 ```
 
-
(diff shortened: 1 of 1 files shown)


dannyhw (approved):
looks good 👍

dannyhw:
gist was also updated

PR #248, opened 2021-08-05
refactor: Refactor class components in addons to function components

### LINKED ISSUE #235: [6.0] Refactor addons to use function components
It would be nice to continue the work that was done on app/react-native to improve fast reload by making all components function components with hooks etc.

lauriharpf:
Issue: #235

## What I did
Fixes #235 by refactoring the class components in all ondevice addons to function components. Note the exceptions mentioned in the comments of commit https://github.com/storybookjs/react-native/pull/248/commits/b671dd626354370ec85f2c34b6375416ca73a874

## How to test
Run the example app at `examples/native` and experiment with the addons.

### DIFF
--- a/addons/ondevice-actions/src/components/ActionLogger/Inspect.tsx
+++ b/addons/ondevice-actions/src/components/ActionLogger/Inspect.tsx
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import React, { useState } from 'react';
 import { Button, View, Text, StyleSheet } from 'react-native';
 
 const theme = {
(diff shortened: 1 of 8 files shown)


addons/ondevice-actions/src/components/ActionLogger/Inspect.tsx
```diff
@@ -20,123 +20,121 @@ const theme = {
...
+  const [expanded, setExpanded] = useState(false);
+  const toggle = (
+    <View style={styles.container}>
+      {name &&
+      ((Array.isArray(value) && value.length) ||
+        (value &&
+          typeof value === 'object' &&
+          !Array.isArray(value) &&
+          Object.keys(value).length)) ? (
+        <Button
```
dannyhw, line 38: Not necessary for this pr I guess but it might be good to replace this button with a custom touchable opacity or pressable to improve the UI for the Android version.

dannyhw (approved):
Not able to test this right now since I'm away without my laptop for a while. However from what I can tell it looks good to me.

Nice job 👍

addons/ondevice-actions/src/components/ActionLogger/Inspect.tsx
lauriharpf: Good idea! Didn't do that yet, but we could tackle it either in this PR or as a separate PR.

PR #250, opened 2021-08-10
docs: Fixes #247 by editing the v6 alpha guide with instructions on how to set up with Expo

lauriharpf:
Also fixes linter warning in README.md.

Issue: #247 

## What I did
Slightly modified the v6 alpha guide so that 6.0 alpha can also be set up on Expo. Notice that there still seem to be some issues with Expo (e.g. the React Native hooks, see https://github.com/storybookjs/react-native/issues/247#issuecomment-897124231)

## How to test
Test the modified v6 alpha guide to ensure that it is understandable & produces a working setup on Expo (and on vanilla React Native, if the user so chooses).

- Does this need a new example in examples/native? **no**, not in my opinion, but we could consider adding an example Expo  project to the repo alongside the vanilla React Native project?
- Does this need an update to the documentation? **included**

PR #251, opened 2021-08-14
fix: Fix #120 (running  "yarn web" on Expo fails).

### LINKED ISSUE #120: react-native-web 0.12.0+ doesn't work with storybook due to deprecated components
**Describe the bug**
In React Native, while i run the command 'yarn web' storybook gives the following warnings:

Attempted import error: 'ViewPropTypes' is not exported from 'react-native-web/dist/index' (imported as 'RNViewPropTypes').

Attempted import error: 'TimePickerAndroid' is not exported from 'react-native-web/dist/index'.

Attempted import error: 'TimePickerAndroid' is not exported from 'react-native-web/dist/index'.

Attempted import error: 'TimePickerAndroid' is not exported from 'react-native-web/dist/index'.

Attempted import error: 'DatePickerIOS' is not exported from 'react-native-web/dist/index'.

Attempted import error: 'DatePickerAndroid' is not exported from 'react-native-web/dist/index'.

Attempted import error: 'AsyncStorage' is not exported from 'react-native-web/dist/index'.

Screenshot from 2020-10-21 15-21-18

And on the browser it gives the error 

**To Reproduce**

expo init appName

npx -p @storybook/cli sb init --type react_native

yarn add @storybook/addon-ondevice-actions @storybook/addon-ondevice-knobs @storybook/addon-actions @storybook/addon-knobs

yarn add @storybook/react-native-server

yarn start-storybook
yarn web

**Code snippets**
In App.js
export {default} from './storybook';

Package.json

{
  "main": "node_modules/expo/AppEntry.js",
  "scripts": {
    "start": "expo start",
    "android": "expo start --android",
    "ios": "expo start --ios",
    "web": "expo start --web",
    "eject": "expo eject"
  },
  "dependencies": {
    "@storybook/addon-actions": "^6.0.26",
    "@storybook/addon-knobs": "^6.0.26",
    "@storybook/addon-ondevice-actions": "^5.3.23",
    "@storybook/addon-ondevice-knobs": "^5.3.23",
    "@storybook/react-native": "^5.3.23",
    "@storybook/react-native-server": "^5.3.23",
    "babel-loader": "^8.1.0",
    "expo": "~39.0.2",
    "expo-status-bar": "~1.0.2",
    "formik": "^2.2.0",
    "react": "16.13.1",
    "react-dom": "16.13.1",
    "react-native": "https://github.com/expo/react-native/archive/sdk-39.0.3.tar.gz",
    "react-native-web": "~0.13.12",
    "yup": "^0.29.3"
  },
  "devDependencies": {
    "@babel/core": "~7.9.0"
  },
  "private": true
}

**System:**

**Additional context**
May be the issue is with the version of react-native-web

lauriharpf:
Issue: Fix #120. The "dist" output in the v6 alpha packages contains JSX and ESNext constructs. Browsers can't handle these natively, so "yarn web" fails on Expo. Fix by using slightly less modern constructs in our dist outputs.

## What I did
Changed targets from `ESNext` to `ES6` (which modern browsers can run without transpiling). Changed the `jsx` output settings to `react` from `react-native` so that the JSX can also be rendered by browsers in addition to React Native. See https://www.typescriptlang.org/docs/handbook/jsx.html#basic-usage .

## How to test
1. Pull this branch
2. Run `yarn prepareAll` in the root of directory of this branch
3. Set up an Expo v6 alpha project using the instructions in https://github.com/storybookjs/react-native/pull/250, but with a small twist: Instead of doing

```shell
yarn add @storybook/react-native@next \
            @react-native-async-storage/async-storage \
            @storybook/addon-ondevice-actions@next \
            @storybook/addon-ondevice-controls@next \
            @storybook/addon-ondevice-backgrounds@next \
            @storybook/addon-ondevice-notes@next \
            @storybook/addon-actions \
            @react-native-community/datetimepicker \
            @react-native-community/slider \
            @storybook/addon-controls
```

point the React Native Storybook dependencies to your local files:

```shell
yarn add /REPLACE-WITH-PATH-TO-MY-REACT-NATIVE-STORYBOOK-REPO/app/react-native \
            @react-native-async-storage/async-storage \
            /REPLACE-WITH-PATH-TO-MY-REACT-NATIVE-STORYBOOK-REPO/addons/ondevice-actions \
            /REPLACE-WITH-PATH-TO-MY-REACT-NATIVE-STORYBOOK-REPO/addons/ondevice-controls  \
            /REPLACE-WITH-PATH-TO-MY-REACT-NATIVE-STORYBOOK-REPO/addons/ondevice-backgrounds \
            /REPLACE-WITH-PATH-TO-MY-REACT-NATIVE-STORYBOOK-REPO/addons/ondevice-notes  \
            @storybook/addon-actions \
            @react-native-community/datetimepicker \
            @react-native-community/slider \
            @storybook/addon-controls
```

4. Run `yarn web` in the root directory of the Expo v6 alpha project that you set up. The project should start up in your browser. If you repeat this with the `next-6.0` branch instead of this branch, `yarn web` should cause an error similar to this:

```
/Users/lauriharpf/Repos/lauriharpf/appName/node_modules/@storybook/react-native/dist/preview/Preview.js 12:14
Module parse failed: Unexpected token (12:14)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| const STORAGE_KEY = 'lastOpenedStory';
| export default class Preview {
>     _clientApi;
|     _storyStore;
|     _addons;
```

5. Experiment with `yarn ios` and `yarn android` in the Expo project as well as `yarn ios` and `yarn android` in the vanilla React Native project at`/examples/native` in this repo. All should work.

### DIFF
--- a/addons/ondevice-controls/tsconfig.json
+++ b/addons/ondevice-controls/tsconfig.json
@@ -1,13 +1,12 @@
 {
   "compilerOptions": {
     "baseUrl": ".",
-    "jsx": "react-native",
+    "jsx": "react",
     "rootDir": "./src",
-    "module": "ESNext",
+    "module": "ES6",
     "skipDefaultLibCheck": true,
     "skipLibCheck": true,
-    "target": "ESNext",
-    "lib": ["ESNext"],
+    "target": "ES2016",
     "allowSyntheticDefaultImports": true,
     "paths": {
       "@emotion/native": ["src/typings.d.ts"]
--- a/app/react-native/tsconfig.json
+++ b/app/react-native/tsconfig.json
@@ -1,13 +1,12 @@
 {
   "compilerOptions": {
     "baseUrl": ".",
-    "jsx": "react-native",
+    "jsx": "react",
     "rootDir": "./src",
-    "module": "ESNext",
+    "module": "ES6",
     "skipDefaultLibCheck": true,
     "skipLibCheck": true,
-    "target": "ESNext",
-    "lib": ["ESNext"],
+    "target": "ES2016",
     "allowSyntheticDefaultImports": true,
     "paths": {
       "@emotion/native": ["src/typings.d.ts"]
--- a/addons/ondevice-notes/tsconfig.json
+++ b/addons/ondevice-notes/tsconfig.json
@@ -1,13 +1,12 @@
 {
   "compilerOptions": {
     "baseUrl": ".",
-    "jsx": "react-native",
+    "jsx": "react",
     "rootDir": "./src",
-    "module": "ESNext",
+    "module": "ES6",
     "skipDefaultLibCheck": true,
     "skipLibCheck": true,
-    "target": "ESNext",
-    "lib": ["ESNext"],
+    "target": "ES2016",
     "allowSyntheticDefaultImports": true,
     "outDir": "dist/",
     "moduleResolution": "node",


addons/ondevice-controls/tsconfig.json
```diff
@@ -1,13 +1,12 @@
...
     "baseUrl": ".",
-    "jsx": "react-native",
+    "jsx": "react",
     "rootDir": "./src",
-    "module": "ESNext",
+    "module": "ES6",
     "skipDefaultLibCheck": true,
     "skipLibCheck": true,
-    "target": "ESNext",
-    "lib": ["ESNext"],
```
dannyhw, line 10: Is lib not needed? What's the default?

app/react-native/tsconfig.json
```diff
@@ -1,13 +1,12 @@
 {
   "compilerOptions": {
     "baseUrl": ".",
-    "jsx": "react-native",
+    "jsx": "react",
```
dannyhw, line 4: What's the effect of this change? I'm wondering if it will now allow dom elements which would be a weird side effect.

dannyhw:
Overall seems like the right move, just a few small questions :)

lauriharpf:
> Overall seems like the right move, just a few small questions :)

Awesome 👍 . Ask away, will try to answer the best I can :)

dannyhw:
> > Overall seems like the right move, just a few small questions :)
> 
> Awesome 👍 . Ask away, will try to answer the best I can :)

I left the questions as review comments.

dannyhw (approved):
Sorry I thought the questions would show up without submitting the review. Overall I think I'm ok with these changes, just clarification on those points would be good 👍.

addons/ondevice-controls/tsconfig.json
lauriharpf: This is a bit complex and I'm not 100% sure that I have understood it correctly, but here's how I think`target` and `lib` work:

* `target`: Defines the JavaScript syntax that is generated (e.g. `ES2016`)
* `lib`: What level of JavaScript is the runtime, _including any polyfills you might bundle in_, expected to support.

If `lib` isn't set, it defaults to the same as the `target` (`ES2016` in this case). The possibility to set a different `target` and `lib` seems to be a way to get around situations where you're including polyfills for some new APIs but want to in general compile down to an older syntax.

https://stackoverflow.com/a/42097465/843984 is a pretty good answer on this. The linked TSConfig docs are also worth a read.

app/react-native/tsconfig.json
lauriharpf: This is another tricky issue, but based on https://www.typescriptlang.org/docs/handbook/jsx.html#basic-usage , it seems that the difference is:

* `react` = transforms JSX syntax into `React.createElement` calls, so the code can run e.g. on the browser that doesn't understand JSX
* `react-native` = leaves JSX syntax as-is (which works when running on React Native, but browsers can't understand it)

There are multiple options and as you say, it is possible that changing this could have some negative side-effects. While I didn't identify any issues while testing, that's of course not a guarantee that we won't run into something. We could perhaps run with this and revert & research more if we notice any undesirable effects.

dannyhw: Ok makes sense I guess there's no harm in changing this.

addons/ondevice-controls/tsconfig.json
dannyhw: The way I understand this is that lib just defines what code you allow in your codebase, such that if you use ESNext then you can use esnext features in the code base but then once you compile it it gets tranformed to the target javascript. So based on that there is no reason to remove the lib option unless we want to change the type of code we write on the project since it wont effect the output.

lauriharpf: Using `"lib": ["ESNext"],` seems to allow using ESNext features which get bundled into the compiled code. One way to test this is to do the following:

1. Change `addons/ondevice-controls/tsconfig.json` to include `"lib": ["ESNext"],` in addition to the `"target": "ES2016",`
1. In `addons/ondevice-controls/src/ControlsPanel.tsx`, add `Promise.any([Promise.reject(1), Promise.resolve(2)]).then((it) => console.log(`resolved ${it}`));` to the component
1. Run `yarn prepareAll` in the project root. It should compile without issues.
1. Open `addons/ondevice-controls/dist/ControlsPanel.js`. The code that uses the ESNext feature `Promise.any` should be included in the compiled source.
1. Set up a project that refers to these local sources. If the runtime does not support `Promise.any` (e.g. Node 14 and below), running the project should fail with 

```
TypeError: Promise.any is not a function. (In 'Promise.any([Promise.reject(1), Promise.resolve(2)])', 'Promise.any' is undefined)

This error is located at:
    in ControlsPanel (created by Wrapper)...
```

Screen Shot 2021-09-12 at 0 09 25

If you remove `"lib": ["ESNext"],` from `addons/ondevice-controls/tsconfig.json` and run `yarn prepareAll` again, then compile should fail with

```
src/ControlsPanel.tsx(47,11): error TS2550: Property 'any' does not exist on type 'PromiseConstructor'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2021' or later.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

dannyhw: So what should we do? go ahead with lib ESNext? Or remove the lib field?

lauriharpf: Would remove the lib and just go with the `target` setting. To keep the codebase modern, we can check how aggressive we could be with the `target` setting without breaking runtime environments. It might be possible to e.g. bump from `ES6`/`ES2015` all the way to `ES2019`, but I haven't tested that.

dannyhw: I've checked out this PR and it seems to work well. I'm happy to merge this and we'll see how it goes and we can update it later if necessary.

dannyhw (approved):
Seems good, the only thing I would say is that we need to apply this to the other addons also. I'll open a new PR with those changes.

Issue #255, opened 2021-08-17
[6.0] resolverMainFields breaks uuid library / crypto polyfill


jd20:
**Describe the bug**
Per #174, it's necessary to set `resolverMainFields: ['sbmodern', 'main']` to get SB 6.0 working with React Native. However, this seems to break projects that depend on `uuid` (probably due to the way `crypto`  module is polyfilled in the `index.js`, see https://github.com/uuidjs/uuid#react-native--expo).

Removing the `resolverMainFields` will fix the `uuid` error on app startup, but obviously that breaks other things in Storybook. I've also tried removing `uuid`, and found `nanoid` is another library with same issue, it's hard to avoid having any dependencies on either of these in a large project.

**To Reproduce**
1. Create a React Native app with SB 6.0.
2. `yarn add uuid react-native-get-random-values`
3. At the beginning of `index.js`, add `import 'react-native-get-random-values'`.
4. Anywhere in `App.tsx`, reference uuid library, for example: `import {v4} from 'uuid'; v4();`
5. Run `npx react-native run-ios`, see error on app startup.
6. Observe the error: `Unable to resolve module crypto from .../RnSBSixAlpha/node_modules/uuid/dist/rng.js: crypto could not be found within the project or in these directories: node_modules`
7. Removing the `resolverMainFields` override from `metro.config.js` will make the error go away.

Here is a link to a minimal example, that reproduces the issue: https://github.com/jd20/RnSBSixAlpha

**Expected behavior**
The app should load without errors.

**Screenshots**
<img width="520" alt="Screen Shot 2021-08-16 at 7 01 27 PM" src="https://user-images.githubusercontent.com/17896438/129651909-5a84a3df-a69d-48c7-a7ed-4c21ba615d11.png">

**System:**
```
  System:
    OS: macOS 11.4
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
  Binaries:
    Node: 15.5.1 - ~/.asdf/installs/nodejs/15.5.1/bin/node
    Yarn: 1.22.10 - ~/.asdf/shims/yarn
    npm: 7.3.0 - ~/.asdf/installs/nodejs/15.5.1/bin/npm
  Browsers:
    Chrome: 92.0.4515.131
    Safari: 14.1.1
  npmPackages:
    @storybook/addon-actions: ^6.3.7 => 6.3.7
    @storybook/addon-controls: ^6.3.7 => 6.3.7
[...]

PR #256, opened 2021-08-19
docs: Update resolverMainFields for vanilla React Native

jd20:
Issue: #255 

## What I did

Add "browser" to the `mainResolverFields`, so as not to break polyfills that depend on it.

An alternative approach would be to override the default config's `mainResolverFields`, but this requires a much bigger change, that I assume some teams may not be comfortable with. Since the default for pretty much all React Native apps should be the same, I think just adding "browser" is a much less invasive change. If a team has customized their `mainResolverFields` already, then it should be pretty clear what they need to do, when updating their `metro.config.js`. Just for reference, below is the altnerative solution:

```ts
const {getDefaultConfig} = require('metro-config');

module.exports = (async () => {
  const {
    resolver: {resolverMainFields},
  } = await getDefaultConfig();
  return {
    transformer: {
      getTransformOptions: async () => ({
        transform: {
          experimentalImportSupport: false,
          inlineRequires: false,
        },
      }),
    },
    resolver: {
      resolverMainFields: ['sbmodern', ...resolverMainFields],
    },
  };
})();
```

## How to test

Run through the v6 instructions, everything should still work same as before.

- Does this need a new example in examples/native? **no**, I did update the existing example though.
- Does this need an update to the documentation? **included**

### DIFF
--- a/examples/native/metro.config.js
+++ b/examples/native/metro.config.js
@@ -17,6 +17,6 @@ module.exports = {
     }),
   },
   resolver: {
-    resolverMainFields: ['sbmodern', 'main'],
+    resolverMainFields: ['sbmodern', 'browser', 'main'],
   },
 };
--- a/v6README.md
+++ b/v6README.md
@@ -49,7 +49,7 @@ yarn add @storybook/react-native@next \
 
 Datetime picker, slider and addon-controls are required for controls to work. If you don't want controls you don't need to install these (controls is the knobs replacement).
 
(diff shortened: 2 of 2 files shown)


dannyhw (approved):
After checking the metro config packages these are the default values anyway and I think its the easier way with fewer changes for the user so I'm ok with this. Maybe we can provide the alternative in the docs somewhere if it becomes necessary. 

Thanks for your contribution 🙇

dannyhw:
gist with the script in has now been updated also

PR #261, opened 2021-09-09
fix(scripts): reduce watcher updates and fix small path bug

dannyhw:
Issue: NA

## What I did

I've made the watcher only update requires when stories are added or removed or the main/preview.js was updated.

I also made an adjustment to how the paths work in the get-stories script.

## How to test

Run the sbn-watcher and sbn-get-stories scripts in the example project.

### DIFF
--- a/app/react-native/scripts/watcher.js
+++ b/app/react-native/scripts/watcher.js
@@ -10,12 +10,20 @@ const {
 
 const log = console.log.bind(console);
 
-const watchPaths = [...getGlobs(), getMainPath()];
+const watchPaths = [getMainPath()];
 if (getPreviewExists()) {
   watchPaths.push(getPreviewPath());
 }
 
-chokidar.watch(watchPaths).on('all', (event, watchPath) => {
-  log(`event ${event} for file ${path.basename(watchPath)}`);
+const updateRequires = (event, watchPath) => {
+  if (typeof watchPath === 'string') {
+    log(`event ${event} for file ${path.basename(watchPath)}`);
+  }
   writeRequires();
-});
+};
+
+chokidar.watch(watchPaths).on('change', (watchPath) => updateRequires('change', watchPath));
+chokidar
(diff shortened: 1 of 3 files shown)


lauriharpf (approved):
Seems to work well :+1: 🚀  🏅 .

Here's how I understand the different operations related to updates, please correct if mistaken:

* `yarn dev` in root React Native Storybook project directory: Causes any changes made to the core React Native Storybook files to be reflected in the example app
* `yarn start` in the examples/native directory: Needs to be running before running `yarn android`/`yarn ios`. When the app is running, changes to existing stories cause a reload & are reflected. New stories do not get added (`storybook.requires.js` only gets written when the app starts, because `prestart` runs `sbn-get-stories`).
* `yarn storybook-watcher` in the examples/native directory: If this is running, then adding/removing stories causes `storybook.requires.js` to be updated, so adding/removing stories is reflected in the app without a restart.

dannyhw:
> Seems to work well 👍 🚀 🏅 .
> 
> Here's how I understand the different operations related to updates, please correct if mistaken:
> 
> * `yarn dev` in root React Native Storybook project directory: Causes any changes made to the core React Native Storybook files to be reflected in the example app
> * `yarn start` in the examples/native directory: Needs to be running before running `yarn android`/`yarn ios`. When the app is running, changes to existing stories cause a reload & are reflected. New stories do not get added (`storybook.requires.js` only gets written when the app starts, because `prestart` runs `sbn-get-stories`).
> * `yarn storybook-watcher` in the examples/native directory: If this is running, then adding/removing stories causes `storybook.requires.js` to be updated, so adding/removing stories is reflected in the app without a restart.

Yeah that's exactly it, the watcher just makes it so you don't need to manually run `sbn-get-stories` after adding a new story its mostly a convenience thing. I'm sure some people would prefer to just run the get stories command instead of having a watcher running.

Issue #247, opened 2021-07-28
[6.0] 6.0.1-alpha.1 does not work with Expo


lauriharpf:
**Describe the bug**
Attempting to follow the [6.0 alpha test instructions](https://github.com/storybookjs/react-native/blob/next-6.0/v6README.md) with Expo causes an error:

```
Failed to compile.
/Users/lauriharpf/Repos/lauriharpf/appName/node_modules/@storybook/react-native/dist/preview/Preview.js 12:14
Module parse failed: Unexpected token (12:14)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| const STORAGE_KEY = 'lastOpenedStory';
| export default class Preview {
>     _clientApi;
|     _storyStore;
|     _addons;
```

**To Reproduce**
Do these steps (slightly modified from the [script version of the 6.0 alpha test instructions](https://gist.github.com/dannyhw/9b84973dcc6ff4fa2e86e32d571d294e)).

```
npm install --global expo-cli
expo init appName
(select blank TypeScript template)
cd appName

yarn add @storybook/react-native@next \
            @react-native-async-storage/async-storage \
            @storybook/addon-ondevice-actions@next \
            @storybook/addon-ondevice-controls@next \
            @storybook/addon-ondevice-backgrounds@next \
            @storybook/addon-ondevice-notes@next \
            @storybook/addon-actions \
            @react-native-community/datetimepicker \
            @react-native-community/slider \
            @storybook/addon-controls

echo "/**
 * Metro configuration for React Native
 * https://github.com/facebook/react-native
 *
 * @format
 */
module.exports = {
  transformer: {
    getTransformOptions: async () => ({
      transform: {
        experimentalImportSupport: false,
        inlineRequires: false,
      },
    }),
  },
  resolver: {
    resolverMainFields: ['sbmodern', 'main'],
  },
};" > metro.config.js;

mkdir .storybook;
mkdir components;
echo "module.exports = {
  stories: [
    './components/**/*.stories.?(ts|tsx|js|jsx)'
  ],
   addons: [
    '@storybook/addon-ondevice-notes',
    '@storybook/addon-ondevice-controls',
    '@storybook/addon-ondevice-backgrounds',
    '@storybook/addon-ondevice-actions',
  ],
};" > .storybook/main.js;

echo "import {withBackgrounds} from '@storybook/addon-ondevice-backgrounds';
export const decorators = [withBackgrounds];
[...]

PR #262, opened 2021-09-11
fix: expo should now work with apha

dannyhw:
Issue: #247 

## What I did

Adjusted the docs so that the steps should now produce a working expo project.

I also added a link to a script that also includes a temporary fix for react-native-web.

## How to test

Follow the steps for expo setup and run the script linked in the docs and test the expo app

### DIFF
--- a/v6README.md
+++ b/v6README.md
@@ -4,7 +4,7 @@
 
 **Vanilla React Native:** Run the [setup script](https://gist.github.com/dannyhw/9b84973dcc6ff4fa2e86e32d571d294e) or follow this guide to do it step-by-step.
 
-**Expo:** Follow this guide.
+**Expo:** Run the [expo setup script](https://gist.github.com/dannyhw/92b3ff0d6ccaead9df2820a507154b87) or follow this guide to do it step-by-step.
 
 This guide assumes that you have yarn, node, npm and npx and they are in your path so your
 terminal can run them. The guide is for Mac/Linux, Windows might need slight adjustments. Contributions for a Windows guide are welcome!
@@ -54,15 +54,16 @@ yarn add @storybook/react-native@next \
 For Expo, we need to separately install the react native packages so Expo can maintain the compatibility for us.
 
 ```shell
-yarn add @storybook/react-native@next \
-            @storybook/addon-ondevice-actions@next \
-            @storybook/addon-ondevice-controls@next \
-            @storybook/addon-ondevice-backgrounds@next \
-            @storybook/addon-ondevice-notes@next \
-            @storybook/addon-actions \
-            @storybook/addon-controls
-
-expo install @react-native-async-storage/async-storage @react-native-community/datetimepicker @react-native-community/slider
+expo install @storybook/react-native@next \
+    @storybook/addon-ondevice-actions@next \
+    @storybook/addon-ondevice-controls@next \
+    @storybook/addon-ondevice-backgrounds@next \
+    @storybook/addon-ondevice-notes@next \
+    @storybook/addon-actions \
+    @storybook/addon-controls \
+    @react-native-async-storage/async-storage \
+    @react-native-community/datetimepicker \
+    @react-native-community/slider
 ```
 
 Datetime picker, slider and addon-controls are required for controls to work. If you don't want controls you don't need to install these (controls is the knobs replacement).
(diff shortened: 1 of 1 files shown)


lauriharpf:
Did some testing, hope I understood correctly:

* `yarn ios` and `yarn android` already work on Expo with the existing tutorial
* `yarn web` doesn't. We can use either this PR or https://github.com/storybookjs/react-native/pull/251 to fix `yarn web` on Expo.

Personally would prefer to avoid adding the `webpack.config.js` changes at https://gist.github.com/dannyhw/92b3ff0d6ccaead9df2820a507154b87#file-setup_expo_rn_sb-sh-L47 to transpile `@storybook` and thus prefer the approach in https://github.com/storybookjs/react-native/pull/251  . There are a couple of reasons for this:

1. `dangerouslyAddModulePathsToTranspile` sounds like something best avoided if possible 🙂 
1. If we can get away with users of Storybook React Native not needing to modify their `webpack.config.js`, it would be great. Requiring to do so adds complexity to the setup. Also, the more customizations we need in `webpack.config.js`, the higher the risk that using Storybook React Native starts to get intertwined with the build customizations the users make (e.g. some combinations don't work).
1. I'm not sure if transpiling solves use of new APIs, such as the example at https://github.com/storybookjs/react-native/pull/251#discussion_r706672892 . Perhaps there's a polyfill that gets added, but haven't tested that.

Don't know if I missed something, but it looks like the tutorial steps in this branch aren't fully in sync with the script at https://gist.github.com/dannyhw/92b3ff0d6ccaead9df2820a507154b87 (the `webpack.config.js` customization is only in the script, not the tutorial?).

One thing I strongly agree with is adding a setup script for Expo, that is an excellent idea :+1: 💪 . We could perhaps include all such setup scripts in this repo itself to make it easier for people to make PRs to adjust them. When we have scripts, we could strip out the `echo "` etc. *nix commands from the tutorial and only include what should be in the file. If someone wants to automate, they can run the scripts fully or partially - the tutorial could be targeted more at those who want to go step-by-step and learn by doing.

dannyhw:
> Did some testing, hope I understood correctly:
> 
> * `yarn ios` and `yarn android` already work on Expo with the existing tutorial
> * `yarn web` doesn't. We can use either this PR or [fix: Fix #120 (running  "yarn web" on Expo fails). #251](https://github.com/storybookjs/react-native/pull/251) to fix `yarn web` on Expo.
> 
> Personally would prefer to avoid adding the `webpack.config.js` changes at https://gist.github.com/dannyhw/92b3ff0d6ccaead9df2820a507154b87#file-setup_expo_rn_sb-sh-L47 to transpile `@storybook` and thus prefer the approach in #251 . There are a couple of reasons for this:
> 
> 1. `dangerouslyAddModulePathsToTranspile` sounds like something best avoided if possible 🙂
> 2. If we can get away with users of Storybook React Native not needing to modify their `webpack.config.js`, it would be great. Requiring to do so adds complexity to the setup. Also, the more customizations we need in `webpack.config.js`, the higher the risk that using Storybook React Native starts to get intertwined with the build customizations the users make (e.g. some combinations don't work).
> 3. I'm not sure if transpiling solves use of new APIs, such as the example at [fix: Fix #120 (running  "yarn web" on Expo fails). #251 (comment)](https://github.com/storybookjs/react-native/pull/251#discussion_r706672892) . Perhaps there's a polyfill that gets added, but haven't tested that.
> 
> Don't know if I missed something, but it looks like the tutorial steps in this branch aren't fully in sync with the script at https://gist.github.com/dannyhw/92b3ff0d6ccaead9df2820a507154b87 (the `webpack.config.js` customization is only in the script, not the tutorial?).
> 
> One thing I strongly agree with is adding a setup script for Expo, that is an excellent idea 👍 💪 . We could perhaps include all such setup scripts in this repo itself to make it easier for people to make PRs to adjust them. When we have scripts, we could strip out the `echo "` etc. *nix commands from the tutorial and only include what should be in the file. If someone wants to automate, they can run the scripts fully or partially - the tutorial could be targeted more at those who want to go step-by-step and learn by doing.

Yeah I actually didn't include it in the tutorial because I only added it in the script for testing purposes, I want to remove it once we have the fix from #120 because I agree with you here that we should work by default where possible instead of including more hacks.

PR #286, opened 2021-10-13
feat: add ability to exclude paths for stories (#285)

### LINKED ISSUE #285: [6.0] option to exclude paths from get stories and the watcher
**Is your feature request related to a problem? Please describe.**

It can be slow in large repos to run the get stories or watcher command and having an exclude option could be useful for avoiding that.

**Describe the solution you'd like**

Something like this would be useful and wouldn't break compatibility with the config file used in other frameworks

```
reactNativeOptions:{
  excludePaths: []
} 
```

**Are you able to assist bring the feature to reality?**
Yes

jgornick:
Adds ability to specify excluded paths in main.js configuration when searching for stories. This helps with glob performance by ignoring node_modules by default.

Fixes #285


app/react-native/scripts/loader.js
```diff
@@ -36,8 +36,18 @@ function writeRequires({ configPath, absolute = false }) {
   const storybookRequiresLocation = path.resolve(cwd, configPath, 'storybook.requires.js');
 
   const main = getMain({ configPath });
+  const reactNativeOptions = main.reactNativeOptions;
+  const excludePaths = reactNativeOptions && reactNativeOptions.excludePaths
```
dannyhw: I think it would be a good idea to split this up a bit, I think 2 ternaries in one assignment is a bit hard to follow.

dannyhw:
@jgornick Thanks for your contribution 🙇 .

Its looking great, if you can make the one change I suggested then I think this is good to go 👍

app/react-native/scripts/loader.js
jgornick: Addressed :)

dannyhw:
Great work!

Issue #308, opened 2021-12-24
Radio control type not working in both Mobile and Web


Karthik-B-06:
**Describe the bug**

Having a radio type control to work with both Mobile and Web but having to control type: 'radio' works on the web and type: 'radios' works on mobile. 

**Expected behavior**
A single control type (either 'radio' or 'radios') working on both Mobile and Web. 

**Screenshots**
![image](https://user-images.githubusercontent.com/35562287/147325078-9f07fa52-85f8-490f-bdea-25df3431e769.png)

**Code snippets**
This code works on mobile. 
```
  argTypes: {
    size: {
      options: <size_options>,
      control: { type: 'radios' },
    },
  },
```
This works for the web. 
```
  argTypes: {
    size: {
      options: <size_options>,
      control: { type: 'radio' },
    },
  },
```

**System:**
```
    "@storybook/addon-ondevice-actions": "^6.0.1-alpha.7",
    "@storybook/addon-ondevice-backgrounds": "^6.0.1-alpha.7",
    "@storybook/addon-ondevice-controls": "^6.0.1-alpha.7",
    "@storybook/addon-ondevice-notes": "^6.0.1-alpha.7",
    "@storybook/addon-react-native-web": "^0.0.14",
```

dannyhw:
@Karthik-B-06 thanks for reporting this issue 🙇‍♂️. If you find any more inconsistencies please let me know.

I'll fix this one I'm back from holidays. Or if you'd like to make a contribution all pr's are welcome.

Happy Holidays 🙂

Karthik-B-06:
Hey, @dannyhw I have made a PR fixing the same.

dannyhw:
This is merged and will be released soon.

PR #309, opened 2021-12-26
fix: fixed inconsistent control type for 'radio'

Karthik-B-06:
Issue: Fixing issue #308 

## What I did
Changed the radio control to work with both Mobile and Web by changing the ControlType string of `radios` to `radio`.

## How to test

- Does this need a new example in examples/native?

No, since the original docs of the storybook include the key with 'radio' 

No, the changes are accommodated in the storybook original docs. 
Screenshot 2021-12-26 at 11 14 29 AM

### DIFF
--- a/addons/ondevice-controls/src/PropField.tsx
+++ b/addons/ondevice-controls/src/PropField.tsx
@@ -14,7 +14,7 @@ type ControlTypes =
   | 'array'
   | 'date'
   | 'button'
-  | 'radios';
+  | 'radio';
 
 export interface Knob {
   name: string;
--- a/addons/ondevice-controls/src/types/index.ts
+++ b/addons/ondevice-controls/src/types/index.ts
@@ -17,5 +17,5 @@ export default {
   select: SelectType,
   date: DateType,
   array: ArrayType,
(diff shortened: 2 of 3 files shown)


dannyhw:
Thanks for your contribution 🙇‍♂️. Will properly review soon. Seems ok to me.

I was thinking it could work for both radios and radio though. 

Also please update the example in the example app :).

Karthik-B-06:
I thought having it work with both radio and radios would create unnecessary confusion. So I chose to use the control type we have in Web. Added the change to Example App too :)

dannyhw:
@Karthik-B-06 ok that makes sense. Looks good. Did you run the example to check it still works? I can't run it right now because I'm away for the holidays. 

Looks good to me though.

Thanks again for contributing :).

If you can just confirm the radio example works on the example app I'm happy to merge it.

Karthik-B-06:
Hey @dannyhw I will get back after checking if the radio control type works on the example app.

dannyhw (approved):
Just tested the example its looking good, thanks for your contribution 🙏 .

Sorry it took so long I've been away for the Christmas break.

Issue #291, opened 2021-10-19
Invalid unit type when insert number in number control


likern:
**Describe the bug**
When I use control type `number` for property
```js
    monthsBefore: {
      control: {
        type: 'number'
      }
    }
```

preview rerenders immediately while I'm typing new value.
That leads to invalid **empty string** being passed to as number value (when I delete old number and try to insert new number).

**Expected behavior**
Don't rerender immediately, only when **Submit** button on keyboard was pressed.

**Screenshots**
![Screenshot_2021-10-19-20-08-48-260_com reactnativeui](https://user-images.githubusercontent.com/842306/137959170-2f1fd2b4-ea9d-4392-9fa7-9dbcc312aa7e.jpg)
![Screenshot_2021-10-19-20-08-02-372_com reactnativeui](https://user-images.githubusercontent.com/842306/137959175-8f0a0e21-15eb-4a09-bdbb-40c3c43b213e.jpg)
![Screenshot_2021-10-19-20-07-56-758_com reactnativeui](https://user-images.githubusercontent.com/842306/137959178-cad9197e-a96a-4f72-ac5e-edc8d7cce187.jpg)

**System:**
```bash
Environment Info:
  System:
    OS: Linux 5.14 Fedora 34 (Workstation Edition) 34 (Workstation Edition)
    CPU: (8) x64 Intel(R) Core(TM) i7-4770K CPU @ 3.50GHz
  Binaries:
    Node: 16.10.0 - ~/.nvm/versions/node/v16.10.0/bin/node
    Yarn: 1.19.0 - ~/.nvm/versions/node/v16.10.0/bin/yarn
    npm: 7.24.0 - ~/.nvm/versions/node/v16.10.0/bin/npm
  Browsers:
    Chrome: 94.0.4606.81
    Firefox: 93.0
  npmPackages:
    @storybook/react-native: ^6.0.1-alpha.6 => 6.0.1-alpha.6
```

dannyhw:
@likern thanks for reporting this. I can see why this is an issue.

I suppose that on blur or similar could be used to defer the update until the user is finished. 

If you want to create a pr with your suggested solution that would be great. If not that's no problem, I'll make sure to include a fix before 6.0.

likern:
@dannyhw Unfortunately, I'm very busy with my duties. I can only report issues I've found while integrating StoryBook. Not able to contribute to this project.

dannyhw:
[...]

PR #315, opened 2022-01-12
fix: enforce number data type for number input (#291)

raychanks:
Issue: #291 

## What I did

Enforce number data type for number input 

## How to test

- Use a number input control in a story
- Change the value of the input to see it's still a number when rendered in the component

Some cases to consider:
- Input with a comma (e.g. 12,3), expect to see the comma replaced with a dot
- Change the input value and then press on the preview section, expect to see the changes reflected
- Input invalid number format (e.g 12...3), expect to see the input border color changes to signify invalid input
- Clear the input will give a zero value
- Put a - sign only will give NaN, while normal negative numbers will give the value of the number

---

- Does this need a new example in examples/native?

No

No

### DIFF
--- a/addons/ondevice-controls/src/ControlsPanel.tsx
+++ b/addons/ondevice-controls/src/ControlsPanel.tsx
@@ -1,6 +1,8 @@
 import styled from '@emotion/native';
 import { API } from '@storybook/api';
-import React from 'react';
+import React, { useState, useCallback } from 'react';
+import { Args } from '@storybook/addons';
+
 import { useArgs } from './hooks';
 import NoControlsWarning from './NoControlsWarning';
 import PropForm from './PropForm';
@@ -46,8 +48,10 @@ export interface ArgTypes {
 const ControlsPanel = ({ api }: { api: API }) => {
   const store = api.store();
   const storyId = store.getSelection().storyId;
+  const [isPristine, setIsPristine] = useState(true);
   const [argsfromHook, updateArgs, resetArgs] = useArgs(storyId, store);
   const { argTypes, parameters } = store.fromId(storyId);
+
   const argsObject = Object.entries(argsfromHook).reduce((prev, [name, value]) => {
     const isControl = Boolean(argTypes?.[name]?.control);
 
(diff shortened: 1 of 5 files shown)


dannyhw:
thanks for the contribution 🙇 

I will review it soon when I have some time.

dannyhw:
sorry for the delay, I've not had the time yet. I haven't forgotten about this though.

dannyhw:
@raychanks hey again sorry for taking so long. 

Seems to solve the issue with an empty input so for the related issue its successful.

The only other thing I noted is that with invalid values the value becomes NaN. Is that what we want? I'm ok with it just wondering what you think about it.

raychanks:
@dannyhw sorry for the late reply. I think that invalid values should be passed in as `NaN` as `NaN` is of type number, and users should be able to test this case. However, one thing I am unsure of is the case of having just a minus sign `-`.

I am thinking about the use case that the user wants to input a negative number. The normal flow would be clearing the input field first, then input the `-` sign, and then type in the number. At the moment of having only the `-` sign, the value would be `NaN`. Should we cater for this case? I think the advantage of handling the `-` sign only would be not having `NaN` transient value, while the disadvantage is that we would have more hidden logic. What do you think?

dannyhw:
@raychanks I think when its `-` we can just set -1 for that special case, I'll push a change for that. I'm going to give this another quick run and then merge. 

Sorry it took so long to get this merged, I've just started a new job and so I have had less time recently.

dannyhw:
I was trying to rebase this and I totally screwed up the entire thing 😩.
edit:
I think I managed to solve it, for a second there I thought it was unrecoverable 😰.

dannyhw:
@raychanks hey I was just checking this one more time and the problem is that the reset doesn't work since the state of 'numStr' is not consistent with the arg state.

Not really sure what the right move is with this now honestly. I'll think on it.

raychanks:
@dannyhw no worries, best of luck to your new role.

Not sure whether a -1 is appropriate, the user might find it more difficult to input other negative numbers like -2. As soon as the user tries to delete the 1 from -1, the input will immediately restore to -1.

Nice catch on the reset bug, I will try to figure out a solution too.

dannyhw:
@raychanks thanks for the kind words :)

in this case the -1 was only set to the Args not to the local state so it didn't effect the typing of the number. Feel free to adjust it though and handle in another way 🙂

raychanks:
@dannyhw you're right about the -1 case.

I added a `isPristine` state to keep track of the reset status. It should be working on the zero value case, reset case and regular number input case now.

dannyhw:
@raychanks interesting change, I'll give it a try later. Looks good though!

dannyhw:
@raychanks this is working great! Great job 🎉 .

Really appreciate your help, and I'm sorry again for taking so long to get to it.

dannyhw (approved):
🚀 🚀 

Great work!

dannyhw:
published this change on 6.0.1-beta.3

Issue #283, opened 2021-10-12
[6.0] select control modal doesn't work correctly on the web


dannyhw:
**Describe the bug**

When using react native web the select control modal doesn't appear correctly. This is because react-native-web doesn't yet support the modal component.

**To Reproduce**
1. run storybook with react native web
2. use a story with a select control
3. go to addons panel and try to use the select control

**Expected behavior**
The select control should provide a normal input of type select when used on the web

**Screenshots**
![image](https://user-images.githubusercontent.com/3481514/137024986-949ef31e-af76-4446-a602-f4f69d7d7943.png)

Issue #281, opened 2021-10-12
[6.0] when ran with react native web the date picker doesn't work


dannyhw:
**Describe the bug**

When using the date control clicking the input has no effect.
A log appears in the console saying that web is not supported for react-native date picker. 

**To Reproduce**
1. run storybook with react native web
2. use a story with a date control
3. go to addons panel and try to use the date control

**Expected behavior**

it should be possible to edit a date

**Screenshots**

![image](https://user-images.githubusercontent.com/3481514/137023916-ef09985d-0609-44ab-888b-7cfffebdd3b6.png)

**Suggested solution**

We can use a regular input with the date and time type when the platform is web

Issue #280, opened 2021-10-12
[6.0] color picker modal doesn't work correctly when ran on web


dannyhw:
**Describe the bug**

The color picker modal is always open and doesn't behave correctly when ran with react-native-web

**To Reproduce**
1. run storybook with react-native-web
2. utilize a story with the color control

**Expected behavior**

A popup or modal should appear when the color square is clicked and then close when the x is pressed

**Screenshots**

![image](https://user-images.githubusercontent.com/3481514/137020794-306de3e6-768a-4c02-9079-8c0e880f5235.png)

raychanks:
Hi Danny, is this issue still persist? I just tried with the web example within the `example/native` folder and I can close the modal.

One issue is the color ring is missing in the web version.

Another issue is when the color picker is opened without changing anything, and then click the Apply button, the applied color is always black. Once the color input is modified and then click the apply button, the selected color is shown. This happens in both web and mobile, not sure whether this is intended.

dannyhw:
@raychanks I guess it could technically be resolved since in 0.14 react native web added modal support. This modal support was added around the time I added this issue so this needs to be re-evaluated.

The color ring might be missing due to the way that the controls package is being built, there is an image used to display it.
Try running `yarn prepareAll` in the root of the mono repo and see if its still broken.

raychanks:
@dannyhw The color ring is still missing after running `yarn prepareAll`. If I import the image resource at the top of the file instead of directly requiring it within `<Image />`, the color ring will be visible. Maybe this is a config issue?

The steps I took is
1. clone the repo
2. yarn
3. yarn prepareAll
4. cd example/native
5. yarn web
![Screenshot 2022-01-19 at 10 19 15 PM](https://user-images.githubusercontent.com/34273644/150149967-0ca3492d-fd3c-4a38-b2e6-637d5040b53c.png)

dannyhw:
@raychanks hmm then maybe some webpack config changes are needed in the example project

dannyhw:
I think I've found a decent solution for this in #322

PR #322, opened 2022-01-31
fix: better web controls

dannyhw:
Issue: #280 #281 #283 

## What I did

- select now uses default select html when on the web platform
- date picker now uses local-datetime input on the web
- color picker now color input type on the web
- some ui improvements for the color picker on device

## How to test

run the example app on web,android and web to check everything is still working

image
image
image

### DIFF
--- a/addons/ondevice-controls/src/types/Color.tsx
+++ b/addons/ondevice-controls/src/types/Color.tsx
@@ -6,6 +6,7 @@ import {
   TouchableOpacity,
   TouchableWithoutFeedback,
   StyleSheet,
+  Platform,
 } from 'react-native';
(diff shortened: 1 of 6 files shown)


raychanks:
The color input seems nicer :)

For the datetime picker, it seems to work fine if I change the value by clicking the calendar icon, but not if I try to type in the values directly. For example, if I want to input 22/01/2022, I cannot input 22 in the day field. Should we also cater for the case of direct input?

dannyhw:
@raychanks I agree its not ideal, though I had some issues making the typing version work. I'm thinking to merge this just to make something work and then add a task to improve it. Thats because currently it just doesn't work on web so something is better than nothing.

PR #354, opened 2022-04-29
Add singleQuote prettier option

gabimoncha:


dannyhw:
I'm not sure about this, I think it would be better to add the file to your prettier ignore since its a generated file and its not possible to match every users prettier config.

dannyhw:
I appreciate the contribution however I think its better to close this because this file (requires.js) shouldn't be edited by the user and therefore should be added to prettier ignore

PR #363, opened 2022-06-28
feat: button feedback for tv users

sebasg0:
Issue:

## What you did
I added user feedback when using tv platforms for react-native-tvos. I will send new prs to improve focus management for tv platforms, so we can test tv components as well.
IMG_4459

## How to test
Use new lib and test feedback out.

- Does this need a new example in examples/native? No

### DIFF
--- a/app/react-native/src/preview/components/OnDeviceUI/navigation/Button.tsx
+++ b/app/react-native/src/preview/components/OnDeviceUI/navigation/Button.tsx
@@ -1,5 +1,5 @@
-import React, { ReactNode } from 'react';
-import { TouchableOpacity } from 'react-native';
+import React, { ReactNode, useState } from 'react';
+import { TouchableOpacity, StyleSheet } from 'react-native';
 import styled from '@emotion/native';
 
 const ActiveBorder = styled.View<{ active: boolean }>(({ active, theme }) => ({
@@ -14,6 +14,12 @@ const ButtonText = styled.Text<{ active: boolean }>(({ theme, active }) => ({
   fontSize: 11,
 }));
 
+const styles = StyleSheet.create({
+  focussedButton: {
+    backgroundColor: 'grey',
+  },
+});
+
 interface Props {
   id: number | string;
   active: boolean;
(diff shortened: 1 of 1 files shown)


dannyhw:
@sebasg0 thanks for the contribution!! 🙇 

👀  nice! I didn't know if it would work on tv or not! 😅 

All PR's welcome!

app/react-native/src/preview/components/OnDeviceUI/navigation/Button.tsx
```diff
@@ -14,6 +14,12 @@ const ButtonText = styled.Text<{ active: boolean }>(({ theme, active }) => ({
   fontSize: 11,
 }));
 
+const styles = StyleSheet.create({
+  focussedButton: {
```
dannyhw, line 18: just a nit pick but I guess it should be focusedButton

app/react-native/src/preview/components/OnDeviceUI/navigation/Button.tsx
```diff
@@ -14,6 +14,12 @@ const ButtonText = styled.Text<{ active: boolean }>(({ theme, active }) => ({
...
+    backgroundColor: 'grey',
```
dannyhw, line 19: I'm wondering if maybe there is some focus styling we can copy from the web storybook, changing the background is ok but probably there is some nice styling defined already for this. I think the web in some cases uses a blue outline.

sebasg0: Will this style be ok?
Screen Shot 2022-06-28 at 10 36 53

sebasg0: got it from web storybook

app/react-native/src/preview/components/OnDeviceUI/navigation/Button.tsx
sebasg0: Oh sure, I missed that one 😅

app/react-native/src/preview/components/OnDeviceUI/navigation/Button.tsx
dannyhw: I think that should be fine, although I guess this is the selected style.

In this screenshot you can see the focused item has only the blue line and the text is grey

image

PR #345, opened 2022-04-11
allow rendering outside safe area

pvinis:
Issue:

## What I did
added support for rendering outside safe area using `parameters: { noSafeArea: true }`.

Tested with `ActionButton`.

With safe area, no nav:
Screenshot 2022-04-11 at 10 37 09

With safe area, with nav:
Screenshot 2022-04-11 at 10 37 21

No safe area, no nav:
Screenshot 2022-04-11 at 10 39 26

No safe area, with nav:
Screenshot 2022-04-11 at 10 39 35

The tiny padding, im not sure where it comes from, but its in the example app, from what i understand.

Please explain how to test your changes and consider the following questions

- Does this need a new example in examples/native?
- Does this need an update to the documentation?

If your answer is yes to any of these, please make sure to include it in your PR.

PR #371, opened 2022-07-15
refactor: some adjustments to safe area styling

dannyhw:
Issue:

## What I did

as a follow up to #345 I've added some changes to make the no safe area settings only apply to the preview and to make it so the navigation bar doesn't float.

This works by having the safe area insets for the navigation bar when the ui is visible but when you hide the ui then the full screen can be used 

## How to test

run the example app and open the outside safe area story

### With UI hidden:

image

### With UI Visible

image


dannyhw:
heres a few small iterations on your changes @pvinis what do you think?

app/react-native/src/preview/components/StoryListView/StoryListView.tsx
```diff
@@ -189,6 +191,7 @@ const StoryListView = ({ selectedStory, storyStore }: Props) => {
           returnKeyType="search"
         />
         <SectionList
+          // contentInset={{ bottom: insets.bottom, top: 0 }}
```
dannyhw, line 192: I'll remove this, forgot to take it out 😅

examples/native/.storybook/preview.js
```diff
@@ -20,5 +20,5 @@ export const parameters = {
 };
 
 const styles = StyleSheet.create({
-  container: { padding: 8, flex: 1 },
+  container: { /* padding: 8, */ flex: 1 },
```
dannyhw, line 23: this is where the padding was coming from btw

PR #379, opened 2022-10-03
fix: make react hooks useable in component story

zhouzh1:
Issue: N/A

## What I did

When I added the `useEffect` and other react hooks in component stories, I encountered the `React has detected a change in the order of Hooks` error, after a investigation, found that the story preview content is got by the method of calling the story function, maybe we need to change it to render the story component. :)

image

## How to test

- Does this need a new example in examples/native? **NO**
- Does this need an update to the documentation? **NO**

### DIFF
--- a/app/react-native/src/preview/components/StoryView/StoryView.tsx
+++ b/app/react-native/src/preview/components/StoryView/StoryView.tsx
@@ -34,9 +34,11 @@ const StoryView = ({ story }: Props) => {
 
   if (story && story.unboundStoryFn) {
     const { unboundStoryFn } = story;
+    const StoryComponent = (context && context.id === story.id) ? unboundStoryFn : null;
     return (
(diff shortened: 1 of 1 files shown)


dannyhw:
@zhouzh1 hey sorry I missed this, seems good to me 👍. I'll just give it a run to make sure but I don't see why it wouldn't work.

zhouzh1:
Hi, @dannyhw, I think it is due to the rendering mechanism inside React, even though a React component is essentially just a function, we still need to let React know that the function is a component, otherwise it will be treated as just a function, or just a part of the hosted component, then the hooks used in the function cause this kind of issue. :)

dannyhw:
@zhouzh1 thanks for the extra info, that's good to know. Although I was referring to your fix seeming like it was fine and I wouldn't find any issue :)

anyway this is live now in the 6.0.1-beta.8 release

PR #380, opened 2022-10-03
fix: Results seemingly not matching filter terms

jonathanj:
In some cases filtering stories gives what appear to be the wrong results, however they are just being rendered with stale views as a result of an overly aggressive `React.memo` equality function.

## What I did

- Removed the `React.memo` equality function, there's no apparent performance impact (in a Storybook with ~100 stories) but the results are now correct

## How to test

https://user-images.githubusercontent.com/112166/193665022-e074f297-1797-4525-af6e-c946390c715c.mov

See the attached video for a reproduction, but the steps are more or less:

- Select the second story in the list, it needs to be anything other than the first story
- Search for "Date", expecting to find the "Date" stories
- See the results appear to be incorrect
- Select the "incorrect" result, and upon re-rendering the correct result is displayed

### DIFF
--- a/app/react-native/src/preview/components/StoryListView/StoryListView.tsx
+++ b/app/react-native/src/preview/components/StoryListView/StoryListView.tsx
@@ -61,9 +61,7 @@ const SectionHeader = React.memo(
       <GridIcon />
       <Header selected={selected}>{title}</Header>
     </HeaderContainer>
-  ),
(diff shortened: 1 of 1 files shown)


dannyhw (approved):

dannyhw:
@jonathanj thanks for your contribution 🙇

PR #397, opened 2022-11-16
fix: add defensive check before call m.hot.accept

nutstick:
Issue: I found that after upgrade to beta.9 will crashed the app in production mode, after debuging found that it's due to m.hot.accept doesn't existed in non dev mode.

## What I did

Add defensive check for `m.hot.accept` before call

## How to test

I forked [expo-template-storybook](https://github.com/nutstick/expo-template-storybook) and update to use latest version.

To reproduce the issue, can run `yarn start --no-dev`. After add the fixes, it can run without crashed.


dannyhw (approved):
makes sense, thanks for the contribution 🙏

Issue #401, opened 2022-11-21
Selected touchable item has a background color of undefined


jgornick:
**Describe the bug**
Using latest 6.0.1-beta.10 (also noticed on 6.0.1-beta.9).

When selecting an item (i.e., story) in the sidebar, the background color isn't being applied, causing the text of the item to blend into the background and look missing.

Based on some some initial research, it seems the [`theme.listItemActiveColor` is undefined](https://github.com/storybookjs/react-native/blob/0288f6d0728b0f29698435b5cd2d46dc7dde57fc/app/react-native/src/preview/components/StoryListView/StoryListView.tsx#L81) for the component.

**To Reproduce**
1. Click on any story in the sidebar list.

**Expected behavior**
The background color should resolve to the theme configured value of "blue-like"

**Screenshots**
Default story selected:
![default-story-selected](https://user-images.githubusercontent.com/8788/203165682-662a4988-cffd-4bae-a39d-bb8d0546f0b1.jpg)

Floating story selected:
![floating-story-selected](https://user-images.githubusercontent.com/8788/203165689-f16c8215-802c-4a4a-a858-b1aaa1b0ef9f.jpg)

Default story selected inspector:
![default-story-background-color](https://user-images.githubusercontent.com/8788/203165698-f88f6756-ffb3-485a-b32f-a8b8f33ae02d.png)

dannyhw:
@jgornick Thanks for reporting this, I haven't seen it myself. Do you know how I can reproduce it?

jgornick:
Hey @dannyhw! Interesting that you haven't seen this before. I didn't think our setup of the StorybookUI was that much different than the example. Here's our [StorybookUI Wrapper](https://gist.github.com/jgornick/0286bc0fcb98d517d518d24234abd688) that we use to initialize the StorybookUI with support for deep linking to a story.

We then just register this wrapper as our component: `AppRegistry.registerComponent(appName, () => StorybookUIWrapper)`

Are there additional steps needed to provide the theme to the StorybookUI component?

dannyhw:
@jgornick 🤔 strange, well I know that there are sometimes issues with the theming so its probably not a difficult fix. In the past I had to put default values every time the theme was used. I think its just another case like that, though I don't have time right now to check it I will look into as soon as I can.

dannyhw:
probably if we just put like
`({ selected, theme }) => (selected ? { backgroundColor: theme.listItemActiveColor ? 'the-blue-color' } : {})`

[...]

PR #404, opened 2022-11-23
fix: default the list item active color if theme value is undefined

jgornick:
Issue: #401


dannyhw (approved):
thanks for the contribution 🙇

PR #426, opened 2023-02-16
chore: fix missing dep in documentation

matinzd:
Issue:

```
node:internal/modules/cjs/loader:988
  throw err;
  ^

Error: Cannot find module '@storybook/core-common'
```

## What I did

Missing `@storybook/core-common` dependency in documentation.

## How to test

- Run `sb-rn-get-stories`.

### DIFF
--- a/MANUAL_SETUP.md
+++ b/MANUAL_SETUP.md
@@ -30,13 +30,13 @@ You may wish to setup everything yourself, you can use the following guide to do
 **Expo**
 
 ```sh
-expo install @storybook/react-native@next @react-native-async-storage/async-storage
+expo install @storybook/react-native@next @storybook/core-common @react-native-async-storage/async-storage
 ```
 
 **React native CLI**
 
 ```sh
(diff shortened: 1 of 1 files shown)


dannyhw (approved):

matinzd:
Even after installing it requires `react-dom` to work. Should it be like this?

```
error: Error: Unable to resolve module react-dom from /Users/matin/work/app/node_modules/@storybook/preview-web/dist/modern/renderDocs.js: react-dom could not be found within the project or in these directories:
  node_modules/@storybook/preview-web/node_modules
  node_modules
  1 | import React from 'react';
> 2 | import ReactDOM from 'react-dom';
```

@dannyhw

dannyhw:
@matinzd thanks for this contribution 🙇 

Yes react-dom is required for one of the storybook dependencies, although its not really used its just a side effect of using `@storybook/preview-web` to re-use more storybook code.

As you can see I haven't gotten around to udpating these things here.

I've been working on the cli init command so these things should be added automatically when using that
What that means is that this command now works: 
`npx sb@next init --type react_native`

PR #428, opened 2023-02-17
fix: improve metro config

matinzd:
Issue: Metro configuration was not extended properly from the default settings.

## What I did

Extend metro config values from default configuration. 

## How to test

- Run `yarn start --reset-cache`
- Should work as expected


dannyhw:
@matinzd this config didn't work because the root of the mono repo needs to be included and sbmodern needs to be the first in the resolver main fields.

I've pushed the fix and this should work now.

dannyhw:
@matinzd this config didn't work because the root of the mono repo needs to be included and sbmodern needs to be the first in the resolver main fields.

I've pushed the fix and this should work now.

dannyhw:
@matinzd this config didn't work because the root of the mono repo needs to be included and sbmodern needs to be the first in the resolver main fields.

I've pushed the fix and this should work now.

dannyhw:
@matinzd this config didn't work because the root of the mono repo needs to be included and sbmodern needs to be the first in the resolver main fields.

I've pushed the fix and this should work now.

dannyhw:
@matinzd this config didn't work because the root of the mono repo needs to be included and sbmodern needs to be the first in the resolver main fields.

I've pushed the fix and this should work now.

dannyhw (approved):
@matinzd thanks for your contribution 🙇 

However please try running the example app before submitting changes

matinzd:
> @matinzd this config didn't work because the root of the mono repo needs to be included and sbmodern needs to be the first in the resolver main fields.
> 
> I've pushed the fix and this should work now.

It seems I did have some caching issues. Thanks for checking.

dannyhw:
@matinzd ah I see, its no problem :). Thanks again.

matinzd:
This is breaking after upgrading to reanimated v3. @dannyhw I think we need to revert back to:

```js
/**
 * Metro configuration for React Native
 * https://github.com/facebook/react-native
 *
 * @format
 */

module.exports = {
  transformer: {
    getTransformOptions: async () => ({
      transform: {
        experimentalImportSupport: false,
        inlineRequires: true,
      },
    }),
  },
  resolver: {
    resolverMainFields: ['sbmodern', 'react-native', 'browser', 'main'],
  },
}

```

PR #429, opened 2023-02-17
Storylist UI usability improvements

jonathanj:
Issue:

The storylist UI can feel cluttered, some contributing factors from my point of view are:

- Confusing and dominated visual hierarchy: sections have large headers but the section itself is not interactive and is only secondary information, large icons but the icons are only a visual hint
- "Wall of text" effect: because the sections run into one another, it creates the visual effect of a sentence without fullstops
- Easy to lose context: when focusing away from the story list (by navigating, for example), coming back takes an uncomfortable amount of time to regain mental context

## What I did

**FYI: This PR builds on https://github.com/storybookjs/react-native/pull/427**

- Resized the icons and section headers to dominate the visual hierarchy less
- Adjust story indentation so that they align with the section, and are easier to scan
- Highlight the entire section that contains the selected story, to help reestablish workflow context
- Separate sections with whitespace to create implicit groupings
- Adjust the search input to look like a search input

Extras:
- Disabled auto-correct / auto-complete on the search input, this generally just causes chaos because your search term (which is often only part of a word) gets auto-corrected to something else
- Made section headers pressable, which just selects the first story

| Before | After |
|--------|-------|
| image | image |

## How to test

- Does this need a new example in examples/native? **No.**
- Does this need an update to the documentation? **I don't think so.**


dannyhw:
this is really nice! I ran it real quick to try it out and it looks really nice and feels a lot better to navigate.

On android I did find some small issue with the search input. its the nexus5 emulator (a very small phone).

https://user-images.githubusercontent.com/3481514/219655169-f8dcc6b2-672e-4740-a2ae-78d2005ebba7.mov

jonathanj:
> On android I did find some small issue with the search input. its the nexus5 emulator (a very small phone).

I force pushed a change a few minutes ago to fix this, I found it too on my simulator but didn't notice it before I created the PR. Sorry about that.

Issue #265, opened 2021-09-13
Top margin not working on some android devices on expo (pixel 4a)


dannyhw:
**Describe the bug**

**To Reproduce**
1. run the expo script from the v6readme
2. setup pixel 4a emulator
2. run `yarn android`
3. preview is partly covered by the notch

**Expected behavior**

Top margin should make all of the preview visible

**Screenshots**

![image](https://user-images.githubusercontent.com/3481514/133073817-c00bcaaf-3c5d-47fe-9649-3a3a4352e8b5.png)

**Code snippets**

changing the app.tsx to the following is a workaround

```
import React from "react";
import { StatusBar, StyleSheet, View } from "react-native";
import StorybookUIRoot from "./.storybook/Storybook";

export default () => (
  <View
    style={[
      StyleSheet.absoluteFillObject,
        { marginTop: StatusBar.currentHeight },
    ]}
  >
    <StorybookUIRoot />
  </View>
);
```

jorgearuv:
I can take this task @dannyhw

dannyhw:
Hey @jorgeruvalcaba are you still looking at this?

jorgearuv:
Hey @dannyhw, sorry; I've been absent here the last few months. But yeah, I'm still looking at it.

wilkuintheair:
I think this is no longer valid:
https://github.com/storybookjs/react-native/blob/next-6.0/app/react-native/src/preview/components/OnDeviceUI/OnDeviceUI.tsx#L40
In my case the problem exists and I'm on Expo project (dev-client) and `global.Expo`, `global.__expo`, `global.__exponent` are all `undefined`.

I think maybe we can just ignore the `exports.IS_EXPO` condition and set the padding for all the Android devices?

And, since on Android the `SafeAreaView` is just a wrapper around the `View`, I guess the issue should be happening on all of the Android devices, right? Are we able to confirm it's just some devices that are affected?

jorgearuv:
I haven't been able to finish with this, feel free to open a PR @wilkuintheair

dannyhw:
Well there were actually changes around safe area in this project so this would need to be tested again.

dannyhw:
@wilkuintheair the problem is that atleast at the time of creating the issue this would only happen on expo so adding the extra padding on android always would cause issues on android (non expo)

fabioatcorreia:
Is this still an issue?  
Wasn't able to replicate it on my end.

dannyhw:
[...]

PR #431, opened 2023-02-21
Fix safe area stories cut off

jonathanj:
Issue:

In stories with `noSafeArea: false` (the default), the available space for story content is cut off at the bottom (going below the navigation bar) even when hiding the navigation UI.

## What I did

I removed the `marginTop` adjustment in the preview area, this seems to not be necessary for `noSafeArea: false` or `noSafeArea: true` stories, the translate-Y adjustment is perfectly fine.

## How to test

I added a new "Usable area" story that has an absolute-filled box, all 4 borders of the box should always be visible and `noSafeArea` should dictate how much of the screen is allowed to be occupied by the story content.

| Before | After |
|--------|------|
| image            |  image |

### DIFF
--- a/app/react-native/src/preview/components/OnDeviceUI/OnDeviceUI.tsx
+++ b/app/react-native/src/preview/components/OnDeviceUI/OnDeviceUI.tsx
@@ -103,6 +103,7 @@ const OnDeviceUI = ({
   const animatedValue = useRef(new Animated.Value(tabOpen));
   const wide = useWindowDimensions().width >= BREAKPOINT;
   const insets = useSafeAreaInsets();
+  const theme: any = useTheme();
   const [isUIVisible, setIsUIVisible] = useState(isUIHidden !== undefined ? !isUIHidden : true);
 
   const handleToggleTab = React.useCallback(
@@ -162,8 +163,8 @@ const OnDeviceUI = ({
   //
   //   4. Storybook UI is not visible, and `noSafeArea` is true: No margin.
   const safeAreaMargins = {
-    marginBottom: isUIVisible ? insets.bottom + navBarHeight : noSafeArea ? 0 : insets.bottom,
-    marginTop: !noSafeArea ? insets.top : 0,
+    paddingBottom: isUIVisible ? insets.bottom + navBarHeight : noSafeArea ? 0 : insets.bottom,
+    paddingTop: !noSafeArea ? insets.top : 0,
   };
   return (
     <>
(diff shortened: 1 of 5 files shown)


jonathanj:
A colleague showed me this issue: https://github.com/storybookjs/react-native/issues/265

Do you think it is related? I’d be happy to test the changes in this PR against that issue, but I’m not sure I know how to reproduce it. Does it require a specific Android simulator on a specific Android version? Does it need to also be an Expo project?

dannyhw:
@jonathanj yeah exactly it only happened on certain emulators on expo, pixel 4a it looks like. However it was fixable on the user side i think by adding a safearea. Though it might be a statusbar height thing, not sure.

The issue you're referring to is really old though.

dannyhw:
I need to check this but I think margin top has to do with a weird jump in the animation that would happen

jonathanj:
> I need to check this but I think margin top has to do with a weird jump in the animation that would happen

Happy to have any input for things I might have missed, but I personally didn't notice any animation jump on Android or iOS.

jonathanj:
> @jonathanj yeah exactly it only happened on certain emulators on expo, pixel 4a it looks like. However it was fixable on the user side i think by adding a safearea. Though it might be a statusbar height thing, not sure.
> 
> The issue you're referring to is really old though.

I couldn't seem to reproduce anything like this, since there have been a number of changes to the safe area since that issue, do you think it's still relevant?

dannyhw:
I'll try it out when I can and double check but it looks good so I'm assuming there won't be any issue.

jonathanj:
> I need to check this but I think margin top has to do with a weird jump in the animation that would happen

@dannyhw I looked into this more carefully and you're right, there is a weird jump/wobble that happens in the animation removing the `marginTop`. I ended up reworking the way the animation/safe area stuff works, and improving the scaled preview view and the animations around it. I think this solves all of the issues I saw reported and about the safe area and the one I was having specifically with elements being cut off.

I also think the changes to the scaled preview make it much clearer with regards to what space the story content occupies on the device screen. See attached video.

Repeated from the commit message:

In order to create more a more stable animation, one that isn't
affected by safe areas popping in/out, a few refactorings were
necessary:

- Remove the use of `SafeAreaView` and instead manually manage the
  margin using the values from `useSafeAreaInsets`
- Use a stable Y-translation for the scaled preview, to ensure it is
  at the same position regardless of safe area presence, or UI
  visibility

Some additional improvements:

- Center the scaled preview horizontally in the available space,
  instead of using a fixed pixel offset (TRANSLATE_X_OFFSET)
- Reduce the number of views and animated views necessary to achieve
  the animation
- Add a shadow to the scaled preview to indicate the device screen
  area in relation to the story content area
- Measure the navigation bar height, instead of using a fixed value
  that is differently incorrect on each platform
- Add comments to code where explaining something not obvious could
  help the next developer


https://user-images.githubusercontent.com/112166/221198072-1c4b94a3-fd43-457c-a67b-136448fe3a90.mp4

dannyhw:
@jonathanj I'm not sure if this is directly because of these changes but the web UI (wide screen) got a horizontal scroll. 

The web support isn't a focus but its nice to have the widescreen version working so you can use an Ipad simulator for example.

actually I checked and on the ipad it seems fine.

dannyhw (approved):
Thanks for your contribution 🙇 . This is really great 💯 .

Issue #181, opened 2021-06-07
refactor components to a lib folder (like the storybook repo)


dannyhw:
I think it would be cool to have the components together they can be re-usable. The addons panel for example would be worth re-using. This should make ondevice addons easier to develop.

in the storybook repo they have a folder like this

lib/components/src/addon-panel/

this is the package @storybook/components

we can do a similar thing for react native

I think there are other things that could later move to this lib folder like the theming

PR #437, opened 2023-03-01
feat: Improve styling in the addons panel

jonathanj:
Issue:

The addons panel (and addons in general) are missing polished UI styling, as well as being inconsistently styled with aspects like differing paddings/margins between various addons.

## What I did

The largest change is restructuring the theme, the goal is to have enough semantic values to meaningfully style the UI well enough for the default Storybook theme, so that users have as much control over their own Storybook's appearance. Once this was done, it was a matter of replacing hardcoded values across the various pieces of UI in the app and addons.

The theme changes are documented in the "Migrating from 5.3x to 6.5.x" section under "Theming" in MIGRATION.md. I think it might be worth writing a separate piece of (non-migration) documentation around theming, what do you think?

Some motivation for the aesthetic changes:
- Make it clearer what the current tab selection is and give the user a hint what is the pressable part of the tab UI
- Increase the contrast between the panel and the various addon inputs/controls
- Balance the whitespace between inputs better, so that the label and input relationship is more intuitive

A goal for the aesthetic changes was to try keep the styling relatively low-key to avoid drawing attention from the story content itself.

There is a fair amount of duplication between addons and the main app, mostly because there's not really a mechanism to reuse components (like inputs and buttons) between addons and the main app. As we discussed, that is probably a task better left for https://github.com/storybookjs/react-native/issues/181 and the duplication can be cleaned up at a later stage.

A summary of the git commit log:

- Fix safe area in the panels by not applying the story's safe area preferences to the Storybook panel UI
- Fix `KeyboardAvoidingView` padding intruding on the app's UI (focus an input and summon the soft-keyboard to see padding creep up from the bottom)
- Rework the theme greatly and remove ad hoc use of values in addons and the Storybook UI
- Tab styling changes (navigation UI, addons list, knob groups)

### Screenshots

| Before | After |
| ------ | ----- |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |

## How to test

The behaviour of Storybook should be identical to how it was before, any significant changes to the behaviour would be a regression. I've added some additional stories to help test aspects such as all of the controls in one place (like the knobs example). There is a commented section in `examples/expo-example/.storybook/Storybook.tsx` with some partial theme overrides.

I removed all the `theme.value || 'some_default'` patterns from the code, as this is a fair amount of maintenance overhead and in my testing I couldn't actually trigger a situation where `theme.value` returned nothing when it ought to have returned something. Because the user's theme is deep-merged with the default theme, there shouldn't be a case of accidentally missing values.


addons/ondevice-controls/src/components/color-picker/HoloColorPicker.js
```diff
@@ -194,7 +196,9 @@ export class HoloColorPicker extends React.PureComponent {
         </View>
         {this.props.hideSliders ? null : (
           <View>
+            <Text style={{ paddingStart: 4, color: '#859499', fontSize: 12 }}>Saturation</Text>
             <Slider value={s} onValueChange={this._onSValueChange} />
+            <Text style={{ paddingStart: 4, color: '#859499', fontSize: 12 }}>Lightness</Text>
```
dannyhw, line 201: I think this should be Brightness

dannyhw (approved):
this is really great, amazing work @jonathanj!

addons/ondevice-controls/src/components/color-picker/HoloColorPicker.js
jonathanj: The two commonly used names are HSL and HSV. [Wikipedia](https://en.wikipedia.org/wiki/HSL_and_HSV) HSB is a thing but is less commonly referred to that way in most software. It's also referred to as [hsl in CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/hsl).

image

dannyhw: ah ok cool

Issue #433, opened 2023-02-24
Testing UI in current navigation (on-device ui) is hard


matinzd:
**Is your feature request related to a problem? Please describe.**
It's an improvement request.

**Describe the solution you'd like**
The current storybook makes UI testing hard since are doing it on device and there is not enough space to see the changes when we open different parts. 

**Describe alternatives you've considered**
Maybe it makes more sense to move addons to a bottom sheet that can be dismissed and see the visual changes at the same time.

**Are you able to assist bring the feature to reality?**
Yes, if I find a free time. I already took a look at the code and I am somehow familiar how it works.

**Additional context**

The actual story page size when we expand addons:

<img width="577" alt="Screenshot 2023-02-24 at 23 50 42" src="https://user-images.githubusercontent.com/24797481/221314166-859e9c4c-5c20-4f09-9c68-05513565018f.png">

dannyhw:
@matinzd hey, thanks for sharing your thoughts on this. The short version is yes I think this is the right direction to go, using a bottom sheet would be better and I'm very happy for any PR with an implementation for this.


The long version is that this is something I've had planned for a while (trying out a bottom sheet for addons). And I worked on something for this a while ago but I've been focused on v6 compatibility so i put it aside. 

All this to say that I'm very open to this idea, however there are a few things to consider.

Ideally we shouldn't add more dependencies that aren't pure javascript (unless there is a really good reason). Reanimated would make this really easy to do but that forces everyone who uses storybook to adopt reanimated and I think that's not what we want.

So that leaves us with either implementing it using the rn animated apis or finding a good package that already does this.

I also want to know how this would change the current UI. Opening a bottom sheet from a tab wouldn't make sense really so where would the bottom sheet get triggered from? If addons isn't a tab anymore then does it make sense to use a tab style navigation at all?

Basically when I think about this in more detail it brings up a lot of questions around how it should work.

How would you go about this? How do you think it should work?

matinzd:
I am also really in favor of not adding more native dependencies so that everyone can use the storybook without adding too much dependencies. 

For bottom sheet we have two options. 

- [react-native-modalize](https://github.com/jeremybarbet/react-native-modalize) (uses rn animated api)
- [@gorhom/bottom-sheet](https://github.com/gorhom/react-native-bottom-sheet) (uses reanimated api)

From a UI perspective, I think we can discuss more on that. I have some ideas but they are not organized.

jonathanj:
[...]

PR #438, opened 2023-03-03
feat: Make it possible to open the addons in bottom split

jonathanj:
Issue: #433 

The addons side panel can be useful but because it scales the preview, it is not a good fit for changing story controls.

## What I did

Make a navigation bar button that toggles displaying the addons panel in a bottom split, this can be controlled with the `isSplitPanelVisible` parameter (like `isUIHidden`) too. I also refactored the UI state for visibility into atoms, so that that state (and its setter) didn't need to be drilled down several levels, and only the relevant UI changes are affected by changes to that state.

When a panel other than the preview panel is being viewed, and the split is visible, it is rendered as a skeleton view. The main reason for this is that I found the tiny scaled version of the addons panel very visually noisy and distracting, and impossible to actually interpret at that size anyway. The skeleton gives a much better transition impression as a result.

I opted to make the split button permanently visible, in my testing it was extremely useful and I think I prefer it for controls. If users want a way to turn it off (for some reason) then I guess that can be added later based on a real-world case.

Summary of the git commits:

- Add a new button to the navigation bar, next to the visibility button, that shows the addons panel in a split view at the bottom of the story preview
- Extract `isUIVisible` state into an atom, so that multiple levels of prop-drilling can be avoided and the `VisibilityButton` can wholly own the state
- Use icons for the navigation bar actions, so that they can be differentiated more easily (the icons are my original creations, so no attribution is necessary, I'm happy to contribute them)
- Refactored the icon code to a single generic `Icon`
- Fix a bug in `useArgs` where `initialState` is not synced when the story changes

https://user-images.githubusercontent.com/112166/222725026-5b3d0913-445a-43fb-939b-2d440c3ca6f8.mp4

## How to test

- Press the split button to toggle the split
- Use the `isSplitPanelVisible` param to control the initial state of the panel
- I added two new background colours to the "Backgrounds CSF" story, black and white, to test the visibility of the icons on those two extremes.


addons/ondevice-controls/src/ControlsPanel.tsx
```diff
@@ -52,19 +52,29 @@ const ControlsPanel = ({ api }: { api: API }) => {
...
-    return isControl
-      ? {
-          ...prev,
-          [key]: { ...argType, name: key, type: argType?.control?.type, value: argsfromHook[key] },
-        }
-      : prev;
-  }, {});
+  const [argsFromHook, updateArgs, resetArgs] = useArgs(storyId, store);
+  const { argsObject, argTypes, parameters } = React.useMemo(() => {
+    const { argTypes, parameters } = store.fromId(storyId) as StoryContext;
```
dannyhw: the variable name is already being used so I'm getting eslint warnings here

dannyhw:
FIrst of all I really love this, it turned out better than I expected. Thanks again for your contributions 🙏.

A bit of a nit pick here is that I'm getting some new eslint warnings, it's only a minor thing but if you think that some of those rules aren't useful we can also just disable them (no inline styles for example).

app/react-native/src/preview/View.tsx
```diff
@@ -173,12 +174,16 @@ export class View {
       }, []);
 
       if (onDeviceUI) {
+        syncExternalUI({
```
dannyhw: Could you explain a bit the context around this? This seems like something that will run every render is that ok?

dannyhw:
One thing that feels a bit weird now with the addons on the bottom is how it doesn't dismiss the keyboard when I press on the preview. I don't think it did before but now it feels more important.

https://user-images.githubusercontent.com/3481514/222759136-28d422c0-37b5-4bc7-9231-199709794452.mov

dannyhw:
I just can't overstate how much better this is 😄, love it!!

addons/ondevice-controls/src/ControlsPanel.tsx
jonathanj: I renamed all the internal variables to avoid the warning about name shadowing.

app/react-native/src/preview/View.tsx
jonathanj: Good catch! So this piece of code was in the wrong place, it was supposed to be done outside of the render, which is where I've moved it to and added a comment about it.

It was relatively harmless inside the render, since the parameters weren't changing and thus the atoms would just have been set to the same values, but doing it every render is definitely not what I intended.

jonathanj:
> One thing that feels a bit weird now with the addons on the bottom is how it doesn't dismiss the keyboard when I press on the preview. I don't think it did before but now it feels more important.

This was such a great piece of feedback, thank you! I tracked this issue down and fixed it. Tested that it didn't interfere with normal presses (like the action button story) or in trying to focus a series of inputs one after the other, seems great now and much nicer to dismiss the keyboard.

dannyhw:
@jonathanj I'm a little unsure about wrapping the story with the touchable component, just because I've done that before and had one or two issues. That being said I also tried it out and it seemed to be working well. Will just keep an eye on that.

PR #440, opened 2023-03-03
fix: StoryView was interfering with scroll gestures

jonathanj:
Some touch interactions, especially scrolling, were not working well after `StoryView` started wrapping its contents in `TouchableWithoutFeedback`.

## What I did

Switch out the use of `TouchableWithoutFeedback` and prefer to use [`onStartShouldSetResponder`](https://reactnative.dev/docs/view#onstartshouldsetresponder) to listen for touch interactions, dismiss the keyboard, and then decline to handle the touch interaction. This means that the interaction attempt is handed to the next component in the hierarchy to handle.

Unlike `Touchable...` which _wants_ to handle the interaction, and ends up interfering with interactions.

## How to test

There are a few new stories under the "Interaction Examples" section in the storybook. They all work now, but you can see the broken behaviour by making the following local modifications in `StoryView.tsx`:

- Remove `onStartShouldSetResponder={dismissOnStartResponder}`
- Wrap the story view content:
  ```
  <TouchableWithoutFeedback onPress={Keyboard.dismiss}>
    <Box …>…</Box>
  </TouchableWithoutFeedback
  ```

Most of the interactions are unaffected, but the "Scrolling" story will not be scrollable unless the keyboard is open.

If you remove both the `TouchableWithoutFeedback` and the `onStartShouldSetResponder` prop, then the scrolling interaction views will dismiss the keyboard on tap, but the "Static" story won't.

### DIFF
--- a/app/react-native/src/preview/components/StoryView/StoryView.tsx
+++ b/app/react-native/src/preview/components/StoryView/StoryView.tsx
@@ -1,8 +1,24 @@
 import React from 'react';
 
-import { Text, TouchableWithoutFeedback, Keyboard } from 'react-native';
+import { Text, Keyboard } from 'react-native';
 import { useStoryContext } from '../../../hooks';
 import { Box } from '../Shared/layout';
+
+/**
+ * This is a handler for `onStartShouldSetResponder`, which dismisses the
+ * keyboard as a side effect but then responds with `false` to the responder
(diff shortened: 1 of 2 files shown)


dannyhw (approved):

dannyhw:
Looks good!

PR #442, opened 2023-03-15
Fix webpack5 and @storybook/addon-storyshots compatibility error, and a typo

leonardo-fc:
## How to test the bug and the fix

```sh
git clone https://github.com/leonardo-fc/patched-react-native-storybook-test.git /tmp/patched-react-native-storybook-test
cd /tmp/patched-react-native-storybook-test
yarn install

# Runs the current @storybook/react-native-server with webpack v4 and v5
# and the patched one with webpack v4 and v5
yarn test:server

# Runs @storybook/addon-storyshots with the current and the patched @storybook/react-native
yarn test:storyshots

# Take a look at the tests files to see how things are being tested
```


dannyhw:
Hey thanks so much for this contribution 🙏.

Just at first glance I'm wondering does the server work with this change for webpack 5? I had some problems when trying this last time.

leonardo-fc:
Updated [patched-react-native-storybook-test](https://github.com/leonardo-fc/patched-react-native-storybook-test) to include start storybook scripts:

```sh
yarn patched-storybook-webpack5:ios
```

https://user-images.githubusercontent.com/124659450/225424134-5e959eab-cffc-4a7e-9806-d1183f616f4b.mp4

dannyhw:
@leonardo-fc this is awesome! Thanks for the video :).

Though I think maybe we should actually use webpack5 by default and only use webpack4 if its specified

leonardo-fc:
Agree
- @expo/webpack-config v18 (Expo SDK 48) already upgraded to webpack5
- storybook v7 no longer have the webpack4 builder

leonardo-fc:
[@storybook/core-server uses webpack5 or webpack4 based on the config file](https://github.com/storybookjs/storybook/blob/v6.5.16/lib/core-server/src/utils/get-manager-builder.ts)
So, should we do something _like_ this? to make webpack5 the default
```js
const mainFilePath = path.join(options.configDir, 'main');

fs.writeFileSync('/somewhere/main.js', `
const config = require(${mainFilePath});

config.core ??= { builder: 'webpack5' };

module.exports = config;
`);

options.configDir = '/somewhere';
```

dannyhw:
@leonardo-fc 

As much as I want to this to work I tried running it on the example project and it fails to run

`Error: Cannot find module '@storybook/manager-webpack-5/manager-preset'`

this file manager-preset isn't in that package.

dannyhw:
@leonardo-fc unfortunately I can't get this to work in the example project in this repository so I can't validate these changes.

Would you be able to get it running with the example project?

dannyhw:
funny enough I was actually able to get it to build moments after sending that...

dannyhw:
@leonardo-fc in future please do your testing in the example project here so that its easier for me to test.

I was however able to get this working and its looking good so thanks for figuring this one out.

dannyhw (approved):

PR #462, opened 2023-04-06
Enable dark mode to use a hook

Jonatthu:
Enables dark mode to use a hook improving the developer experience when changing dark mode or light mode

https://user-images.githubusercontent.com/3742176/230413579-c6ce47b0-c3f7-4667-bd9d-fc201a0de522.mov


dannyhw (approved):
Looks good thanks!

Issue #473, opened 2023-05-06
Unable to resolve module ./storybook.requires


Brandon-Perry:
**Describe the bug**
When I yarn start on my project after installing storybook, I get the following error:

error: Error: Unable to resolve module ./storybook.requires from /Users/brandon/Desktop/monorepo/client/.storybook/index.js: 

None of these files exist:
  * .storybook/storybook.requires(.native|.ios.js|.native.js|.js|.ios.jsx|.native.jsx|.jsx|.ios.json|.native.json|.json|.ios.ts|.native.ts|.ts|.ios.tsx|.native.tsx|.tsx)
  * .storybook/storybook.requires/index(.native|.ios.js|.native.js|.js|.ios.jsx|.native.jsx|.jsx|.ios.json|.native.json|.json|.ios.ts|.native.ts|.ts|.ios.tsx|.native.tsx|.tsx)
  1 | import { getStorybookUI } from '@storybook/react-native';
  2 |
> 3 | import './storybook.requires';
    |         ^
  4 |
  5 | const StorybookUIRoot = getStorybookUI({});
  6 |
    at ModuleResolver.resolveDependency (/Users/brandon/Desktop/monorepo/client/node_modules/metro/src/node-haste/DependencyGraph/ModuleResolution.js:137:15)
    at DependencyGraph.resolveDependency (/Users/brandon/Desktop/monorepo/client/node_modules/metro/src/node-haste/DependencyGraph.js:260:43)
    at Object.resolve (/Users/brandon/Desktop/monorepo/client/node_modules/metro/src/lib/transformHelpers.js:177:21)
    at Graph._resolveDependencies (/Users/brandon/Desktop/monorepo/client/node_modules/metro/src/DeltaBundler/Graph.js:432:35)
    at Graph._processModule (/Users/brandon/Desktop/monorepo/client/node_modules/metro/src/DeltaBundler/Graph.js:218:38)
    at async Graph._addDependency (/Users/brandon/Desktop/monorepo/client/node_modules/metro/src/DeltaBundler/Graph.js:314:20)
    at async Promise.all (index 5)
    at async Graph._processModule (/Users/brandon/Desktop/monorepo/client/node_modules/metro/src/DeltaBundler/Graph.js:263:5)
    at async Graph._traverseDependenciesForSingleFile (/Users/brandon/Desktop/monorepo/client/node_modules/metro/src/DeltaBundler/Graph.js:207:5)
    at async Promise.all (index 0)

[...]

PR #483, opened 2023-06-08
docs: downgrading setup intructions to stable version

flexbox:
Issue: --

## What I did

As a developer
I wanted to install storybook to an existing expo project
So that i can sparks joy in my UI

## Why this PR?

Running the last version will install the last version `@next`

```
❯ npx sb@next init --type react_native
Need to install the following packages:
  sb@7.1.0-alpha.29
Ok to proceed? (y) y
```

this can lead to random issues like #473 

That's why we could ask to install the last stable version

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ npx react-native init MyApp --template react-native-template-storybook
 Run init to setup your project with all the dependencies and configuration files:
 
 ```sh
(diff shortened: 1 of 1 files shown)


dannyhw:
thank you @flexbox  🙇

PR #457, opened 2023-04-02
Storybook React Native - Add Categories matching web

Jonatthu:
Issue: #12 

## What I did

Adds categories/folders for the story list

## How to test

Run the example app and navigate to the new example story

Issue #205, opened 2021-07-11
[6.0] allow for organising stories by category in Story-list 


dannyhw:
In the web version on storybook it is easy to organise stories by type and we should attempt to implement this here

dannyhw:
#338 is the same and has more details

dannyhw:
please try the new alpha version 6.5.5-alpha.0 which has an implementation of this

PR #486, opened 2023-06-18
feat: nested grouping of stories with "group/story" syntax 

dannyhw:
Issue: #205 

## What I did

thanks to @Jonatthu for getting this started with #457

This adds an initial version of categories/nested stories in the sidebar

## How to test

- run example project and see the sidebar and storylist story

https://github.com/storybookjs/react-native/assets/3481514/b2c29fd8-972a-4667-8553-9de33dfaf0e8


dannyhw:
@SocketSecurity ignore telejson@7.1.0

packages/react-native/src/hooks.tsx
```diff
@@ -41,11 +42,31 @@ export function useIsStorySelected(storyId: string) {
...
+export function useIsChildSelected(entries: StoryIndexEntry[]) {
+  return useAtomValue(
+    useMemo(
+      () =>
+        atom((get) => {
+          const contextId = get(storyContextAtom)?.id;
+          return !!entries.find(({ id }) => id === contextId);
+        }),
+      [entries]
+    )
```
jonathanj, line 69: It's not super clear from reading the calling code whether `item.stories` is a stable reference. Does it only change when Storybook core needs to update (because of HMR or whatever else)?

jonathanj (approved):
I just have two questions, one of which I commented on the changes for. I _think_ that `item.stories` changes only when necessary instead of being recomputed on render, but thought I would ask anyway. The app in the video still looks snappy though, so there's probably not a serious performance regression, feel free to resolve my comment.

My other question is more of a (hypothetical) UX question: What happens if you have a few levels of nesting and medium length story names? Will list items end up squished towards the right? Might be easier to see what problems people run into and address them when/if they happen.

Otherwise this looks like a really nice improvement!

packages/react-native/src/hooks.tsx
dannyhw: When filtering the tree can be recreated, its based on the 'data' state which gets set when filtering

dannyhw:
added flex-srhink and maximum lines for story headers so we get elipsis when it goes over

dannyhw:
Going to merge and make a new alpha version with these changes.

PR #528, opened 2023-11-21
make storage configurable

patlux:
Issue:

I don't want to install and maintain an additional dependency like `@react-native-async-storage/async-storage` when I use another storage library or don't use one at all.

So the storage should be configurable instead hardcoded.

## What I did

I removed `@react-native-async-storage/async-storage` completely as dependency and added an `storage`-option to the params parameter of `getStorybookUi`. So if you want to use AsyncStorage you need to add it like the following example shows:

```tsx
import AsyncStorage from '@react-native-async-storage/async-storage';

const StorybookUIRoot = getStorybookUI({
  storage: {
    getItem: AsyncStore.getItem,
    setItem: AsyncStore.setItem
  }
})
```

Or if you use `react-native-mmkv`:

```tsx
import {MMKV} from 'react-native-mmkv'

const storage = new MMKV()

const StorybookUIRoot = getStorybookUI({
  storage: {
    setItem: async (key, data) => storage.set(key, data),
    getItem: async (key) => storage.getString(key) ?? null,
  }
})
```

## How to test

Don't have the time at the moment to update the documentation, but I will later.

### DIFF
--- a/examples/expo-example/package.json
+++ b/examples/expo-example/package.json
@@ -15,7 +15,6 @@
     "web": "expo start --web"
   },
   "dependencies": {
-    "@react-native-async-storage/async-storage": "1.18.2",
     "@react-native-community/datetimepicker": "7.2.0",
     "@react-native-community/slider": "4.4.2",
     "@storybook/addon-actions": "^6.5.14",
--- a/packages/react-native/package.json
+++ b/packages/react-native/package.json
@@ -81,7 +81,6 @@
     "typescript": "^5.1.3"
   },
   "peerDependencies": {
-    "@react-native-async-storage/async-storage": ">=1",
     "react": "*",
     "react-native": ">=0.57.0",
     "react-native-safe-area-context": "*"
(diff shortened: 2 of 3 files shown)


stevoland:
+1 for this in v7. I have a similar patch in my project.

dannyhw:
Thanks for this contribution, I'm currently travelling so I haven't had time to look through this but i will soon. At first glance it seems good. I just want to consider ways to make this as easy as possible for users since its a key user experience feature to persist selection.

dannyhw:
Haven't forgotten about this, should be able to review this week.

dannyhw:
I thought more about this and I think the best time to introduce a breaking change like this will be in v7 since its just around the corner anyway

dannyhw:
I'm not able to push to your branch in order to resolve conflicts with the v7 branch. You need to enable edits from maintainers. I will add the change manually and give you credit.

patlux:
@dannyhw The checkbox is already checked. You should be able to push

IMG_2511

dannyhw:
I cherry picked this commit into the v7 branch, thanks for your contribution 🙏

dannyhw:
@patlux weird, it said I don't have permission, anyway in the cherry pick your commit is attributed to you.

PR #530, opened 2023-12-17
feat/server v7

dannyhw:
Issue:

## What I did

starting to get the server working

- need to find a way to set globals from the server to the browser with the new v7 architecture


dannyhw (commented):
I just merged this because I accidently included other work in here but there will be more work needed to get the server working still

PR #543, opened 2024-01-26
Failing case for render error in ondevice addons with many stories (v7)

stevoland:
Issue:

Large storybook hits render error in ondevice addon panels eg

```
Error: Cannot call fromId/raw() unless you call cacheAllCSFFiles() first.

This error is located at:
    in Notes (created by Wrapper)
    in RCTScrollContentView (at ScrollView.js:1732)
```

I think a guard something along these lines is required https://github.com/reg-viz/storycap/pull/605/files#diff-bad46b72635dc4ead145dcbaaf545cd2ecd5695f8d83271cf3327237d5368060

## What I did

Hit this error with my actual project. Took a while to track it down to the size of the store.

## How to test

Run storybook dev, go to controls/notes/backgrounds panel on device

### DIFF
--- a/examples/expo-example/.storybook/main.js
+++ b/examples/expo-example/.storybook/main.js
@@ -6,6 +6,11 @@ module.exports = {
       files: '**/*.stories.?(ts|tsx|js|jsx)',
       titlePrefix: 'OtherComponents',
     },
+    {
+      directory: '../many_components',
+      files: '**/*.?(ts|tsx|js|jsx)',
+      titlePrefix: 'ManyComponents',
+    },
     // {
     //   directory: '../other_components',
     //   files: '**/*.storiesof.?(ts|tsx|js|jsx)',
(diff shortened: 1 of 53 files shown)


dannyhw:
Interesting, we actually solved a similar issue in the v8 branch but not for large lists of stories just the `cacheAllCSFFiles() first` thing. The strange thing here though is that our story imports are synchronously evaluated in v7 so I don't understand why this would happen.

dannyhw:
thanks for the details @stevoland and debugging this, I'll look into it this weekend.

dannyhw:
@stevoland Hey we just pushed a fix for this problem can you install `7.6.10-alpha.3` and see if it works for you?

stevoland:
@dannyhw Tested alpha.3 on my project and looks like all is good 🙏 

I think that's the last blocker for me. Thanks for all the hard work!

dannyhw:
@stevoland fyi we now have 7.6.10-rc.0 out as an a release candidate, theres a few more fixes in there including support for main.ts

dannyhw:
closing since the fix was already merged

PR #567, opened 2024-04-12
fix: types was included in the `.js` file output using the `useJs` option in `sb-rn-get-stories`

grgr-dkrk:
Issue: none

## What I did
The `.js` file output using the `useJs` option with `sb-rn-get-stories` contained types.
This PR fixes that.

The type of args of `importFn` is defined in [@storybook/preview-api](https://github.com/storybookjs/storybook/blob/v7.6.17/code/lib/types/src/modules/story.ts#L34), so I deleted the definition in this file.

## How to test
I verified that there are no type errors in the output `.js` file.


dannyhw:
Thanks for the contribution will get to this soon :)

dannyhw:
whilst it does work without the type I added this type specifically because some stricter typescript settings would complain about the type being any. I've adjusted it to make it not generate that type if using js and added tests for js generation.

dannyhw:
thanks again for raising this issue and your contribution :)

dannyhw:
@grgr-dkrk just published this on  7.6.18 please try it :)

PR #570, opened 2024-04-26
automatically strip play functions for react-native runtime

awinograd:
Issue:

## What I did

Play functions are not supported on native. Instead of requiring consumers to guard their play functions with platform checks, we automatically strip any stories of their play functions for them.

## How to test

This PR doesn't fully make writing play functions compatible with @storybook/react-native. See https://discord.com/channels/486522875931656193/490822534233849873/1233464353907540081.

I've worked around the build issue with the following metro config for now which is a bit hacky

```
      resolveRequest: function (context, moduleName, platform) {
        if (moduleName === 'os' || moduleName === 'tty') {
          return context.resolveRequest(context, 'lodash/noop', platform);
        }
```

I don't think it yet makes sense to add an example test showing that the play function is stripped from a story.

Please explain how to test your changes and consider the following questions.

### DIFF
--- a/packages/react-native/src/Start.tsx
+++ b/packages/react-native/src/Start.tsx
@@ -79,7 +79,23 @@ export function prepareStories({
               tags: ['story'],
             };
 
-            importMap[`${root}/${filename.substring(2)}`] = req(filename);
+            const importedStories = req(filename);
+            const stories = Object.entries(importedStories).reduce(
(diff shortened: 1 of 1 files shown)


dannyhw:
@awinograd thanks for the contribution, seems to me like a reasonable solution. Don't have the time right now to test but will do soon.

PR #573, opened 2024-05-03
Update prepareStories to handle more options & strip play functions

awinograd:
Issue:

## What I did

Adds tests for https://github.com/storybookjs/react-native/pull/570 and adds support for includeStories/excludeStories CSF options on meta.

## How to test

Unit tests have been added to cover new/changed functionality

### DIFF
--- a/examples/expo-example/.storybook/main.ts
+++ b/examples/expo-example/.storybook/main.ts
@@ -22,6 +22,9 @@ const main: StorybookConfig = {
     '@storybook/addon-ondevice-backgrounds',
     '@storybook/addon-ondevice-actions',
   ],
+  reactNative: {
+    playFn: false,
+  },
 };
 
 export default main;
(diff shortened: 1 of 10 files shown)


dannyhw:
I'd like to make an adjustment to enable the play fn via an option in the config but this looks good to me.

awinograd:
@dannyhw is it actually possible to execute the play fn on native? If so, I definitely see how the option would be useful (and would also be interested myself in how to enable play fn support). I imagine there must be some significant work in terms of getting `@storybook/test` to play nicely w/native to unblock though.

dannyhw:
@awinograd the play function gets called and you can run javascript in there, we just haven't got a way to query elements at runtime without a tool like detox. The idea is that later we could find a way to make it work, I just don't want to change behaviour right now in a minor version even if it doesn't really do much.

I can make the adjustment it will probably just look like this in your main.js

```
reactNative: {
  playFn: false
}
```

dannyhw (commented):
@awinograd I pushed some changed the pr that adds some options to main.ts that can be expanded on later. Also probably we don't need the other pr since this one has both sets of changes right? Let me know if you're ok with these changes.

awinograd:
@dannyhw the new option looks ok to me! I've closed the other PR since you are right that this one includes that functionality.

Thanks for taking a look at this PR!

dannyhw (approved):

PR #587, opened 2024-07-23
fix: CalledPreviewMethodBeforeInitializationError (8.2)

stevoland:
Fix this error in 8.2-alpha1 with large number of stories:

```
ERROR [SB_PREVIEW_API_0005 (CalledPreviewMethodBeforeInitializationError): Called `Preview.loadStory()` before initialization.

The preview needs to load the story index before most methods can be called. If you want
to call `loadStory`, try `await preview.initializationPromise;` first.

If you didn't call the above code, then likely it was called by an addon that needs to
do the above.]
```

## What I did

Add waits for initialization

## How to test

Start with large number of stories, eg: https://github.com/stevoland/react-native/commit/aca843bc3ead9ee3669349c2e6f62d1eee228078

### DIFF
--- a/examples/expo-example/.storybook/index.tsx
+++ b/examples/expo-example/.storybook/index.tsx
@@ -7,7 +7,7 @@ const StorybookUIRoot = view.getStorybookUI({
     getItem: AsyncStorage.getItem,
     setItem: AsyncStorage.setItem,
   },
-  // enableWebsockets: true,
+  enableWebsockets: true,
 
   // initialSelection: { kind: 'TextInput', name: 'Basic' },
   // isUIHidden: true,
--- a/examples/expo-example/babel.config.js
+++ b/examples/expo-example/babel.config.js
@@ -1,7 +1,22 @@
+// const path = require('path');
+
 module.exports = function (api) {
   api.cache(true);
   return {
     presets: ['babel-preset-expo'],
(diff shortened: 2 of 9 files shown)


dannyhw:
thanks @stevoland 🙏

dannyhw:
seem to be getting a build error here

 Error: src/View.tsx(151,25): error TS2445: Property 'storeInitializationPromise' is protected and only accessible within class 'Preview<TRenderer>' and its subclasses.
  Error: src/View.tsx(183,21): error TS2445: Property 'storeInitializationPromise' is protected and only accessible within class 'Preview<TRenderer>' and its subclasses.
  
  I wonder if theres another way to call it.

stevoland:
Bah, sorry didn't get the tests to run locally.

The error message says initializationPromise but that's undefined. I'll have a dig around

stevoland:
I guess we can just add a subclass and expose it https://github.com/storybookjs/react-native/blob/feat/v8/packages/react-native/src/Start.tsx#L211

stevoland:
Hmm, actually leave this for now. There's a ready() method on preview to use but also I get another early initialisation error when web socket is enabled so needs more work

stevoland:
Ok, I think this does the job now, thanks

dannyhw:
sorry for not following up earlier I've had a lot going on, taking another look now

dannyhw:
seems to be working, thanks for the fix 🙏

PR #591, opened 2024-08-31
chore: fix workspace dev script

fabioatcorreia:
Issue:
Running `yarn dev` on the project's root repository was getting stuck on the first long-running process.  
This prevented from getting automatic feedback when applying changes to some packages.

## What I did

Add `parallel` flag to lerna run, allowing multiple long-running processes 

Reference: https://github.com/lerna/lerna/tree/main/libs/commands/run#--parallel

Also added the `stream` flag to better identify which package the output was coming from.

## How to test

Run `yarn dev` on the project's root repository - a process for all project's workspaces should start

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   },
   "scripts": {
     "bootstrap": "node ./scripts/bootstrap.js",
-    "dev": "lerna run dev",
+    "dev": "lerna run dev --stream --parallel",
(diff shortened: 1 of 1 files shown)


dannyhw:
nice, thanks for the contribution 🙇

dannyhw (approved):

PR #602, opened 2024-09-21
Use theme to populate the markdown styles

trajano:
Issue: only the text was converted to the theme color, this ensures all other elements are switched to the theme color

## What I did

useMemo against the theme default color to create a markdown stylesheet and applied that to the markdown element

## How to test


dannyhw (approved):
made some small adjustments but looks like we're good, thanks for the contribution 🙏

PR #623, opened 2024-10-16
Add core config

trajano:
This is needed to disable telemetry

Issue:

## How to test

### DIFF
--- a/packages/react-native/src/index.ts
+++ b/packages/react-native/src/index.ts
@@ -1,4 +1,4 @@
-import type { StorybookConfig as StorybookConfigBase } from '@storybook/core/types';
+import type { CoreConfig, StorybookConfig as StorybookConfigBase } from '@storybook/core/types';
 import type { ReactNativeOptions } from './Start';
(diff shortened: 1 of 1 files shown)


dannyhw:
I think there might be some extra steps needed to make the config actually apply since this config for rn storybook is parsed manually by the generate script. Have you confirmed that telemetry is going out? And is it being turned off by this option?

trajano:
Nope, I just saw Telemetry config in the documentation while I was doing the update and I was just expecting the typings for the config to be relatively the same.  I am primarily putting it to disable the telemetry in case it is doing telemetry.

dannyhw:
I will check this but it was my understanding that telemetry is not going out since its not explicitly implemented for rn storybook

trajano:
Awesome, it would be nice to know in the documentation that it is not supported by RN storybook in case people have privacy concerns.

dannyhw:
I'm fairly confident that no telemetry is going out from react native and this config would have no impact. If I find anything that contradicts that I will revisit this. Thanks for bringing it up though!

PR #627, opened 2024-10-22
fix: overflowing story content overlaps sidebar

tlow92:
Issue:
If the content of a story overflows the container it might overlap with the sidebar, preventing interactions with it.
image

## What I did
Set `overflow: hidden` to the content container.

image

For demonsteration purpose I changed the Button styles, but it's not part of the PR. @dannyhw Let me know if I should add an example story to demonstrate this.


dannyhw:
hey @tlow92 thanks for the contribution 🙏 

can you add an example to the example project which can show the problem/fix to make sure there is no regression

dannyhw:
I noticed theres actually a similar issue on mobile, pushed a similar change for that layout

dannyhw (approved):
Looks great, thanks 👍

PR #632, opened 2024-10-25
feat: add buttons to hide/show ui

dannyhw:
Issue:

## What I did

image

image

## How to test


packages/react-native-ui/src/icon/FullscreenIcon.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+import Svg, { Path, SvgProps } from 'react-native-svg';
+import { useTheme } from '@storybook/react-native-theming';
+
+export function FullscreenIcon({
+  color, //= '#2E3438',
```
tlow92: I guess this is a missed debug statement?

packages/react-native-ui/src/Layout.tsx
```diff
@@ -133,7 +137,29 @@ export const Layout = ({
...
+            position: 'absolute',
+            bottom: uiHidden ? 56 + insets.bottom : 16,
+            right: 16,
+            backgroundColor: 'white',
+            padding: 4,
+            borderRadius: 4,
+            borderWidth: 1,
+            borderColor: theme.appBorderColor,
+          }}
+          onPress={() => setUiHidden(!uiHidden)}
```
tlow92: Even though we don't have any other calls of `setUiHidden` so it should not be a problem. I would suggest using function setter for toggling.

```suggestion
          onPress={() => setUiHidden(prev => !prev)}
```

packages/react-native-ui/src/Layout.tsx
```diff
@@ -133,7 +137,29 @@ export const Layout = ({
...
   return (
     <View style={{ flex: 1, paddingTop: insets.top, backgroundColor: theme.background.content }}>
-      <View style={{ flex: 1, overflow: 'hidden' }}>{children}</View>
+      <View style={{ flex: 1, overflow: 'hidden' }}>
+        {children}
+
+        <TouchableOpacity
+          style={{
+            position: 'absolute',
+            bottom: uiHidden ? 56 + insets.bottom : 16,
```
tlow92: I think in the future it might be useful to make it draggable (like android notification bubbles), in case it overlaps with buttons of a story component.

packages/react-native-ui/src/Layout.tsx
```diff
@@ -133,7 +137,29 @@ export const Layout = ({
...
+            borderRadius: 4,
+            borderWidth: 1,
+            borderColor: theme.appBorderColor,
+          }}
+          onPress={() => setUiHidden(!uiHidden)}
+        >
+          {uiHidden ? (
+            <CloseFullscreenIcon color={theme.color.mediumdark} />
+          ) : (
+            <FullscreenIcon color={theme.color.mediumdark} />
```
tlow92: Probably it's better to use `theme.background.content`. 
Otherwise it does not respeect browsers darkmode, right?

dannyhw:
thanks for all the feedback @tlow92 I definitely apply those and iterate a bit further to see what I can come up with

dannyhw:
I think this is in a good place now, it would be cool to move it around but I'm thinking of adding that later

To hide this button you can use the following parameter, either globally in preview or at a story level

`hideFullScreenButton: true,`

PR #633, opened 2024-10-26
migrate withStorybook to typescript

### LINKED ISSUE #630: incorrect typing for withStorybook
Thanks for adding the typing but you did it incorrectly at the end

https://github.com/storybookjs/react-native/blob/next/packages/react-native/metro/withStorybook.d.ts

you had 

```
export function withStorybook(config: InputConfigT, options?: WithStorybookOptions): InputConfigT;
```

when it should've been

```
declare function withStorybook(config: InputConfigT, options?: WithStorybookOptions): InputConfigT;

export = withStorybook
```

tlow92:
## Proposal to fix https://github.com/storybookjs/react-native/issues/630

> [!NOTE]
> Instead of manually defining the `d.ts` file I converted the current implementation to typescript and let `tsup` emit the `.js` and `.d.ts` file. This should hopefully be more future-proof.
>
> In general it works, but I had to add a custom tsup cli cmd to the package.json (reason is explained below). 

* Currently `withStorybook` is imported like this `require('@storybook/react-native/metro/withStorybook');`
I think ideally `withStorybook` would live in src like all other files and be transpiled to `dist`, but that would be a breaking change.
* I tried to emit the files to `dist` and put barrel files in `metro` folder so that it is not a breaking change, but that created problems with tsup because apparently it can not handle barrel `.d.ts` files and always emits an empty `.d.ts`.
* The other approach is what I did now:
Using a separate `tsup` cmd to specify a different `outDir` (`metro` folder)


dannyhw:
thanks for getting this underway! I think we got a good solution here. I'm going to merge and start working toward a 8.4 pre-release with multiple changes in it

trajano:
I looked at the code the hack I did with `export = withStorybook` is not needed if you converted to typescript instead just `export function withStorybook` in an earlier line.  

The hack was needed because of the d.ts nature which no longer applies.

PR #638, opened 2024-11-01
fix: add missing devDependency(ts-dedent)

duguyihou:
Issue:

## What I did

add `ts-dedent` to devDependency

## How to test

`ts-dedent` is not in the devDependency list but it is required in `StoryHash.ts`

storybookdemo

### DIFF
--- a/packages/react-native-ui/package.json
+++ b/packages/react-native-ui/package.json
@@ -54,7 +54,8 @@
     "jest": "29.7.0",
     "react-test-renderer": "18.2.0",
     "tsup": "^7.2.0",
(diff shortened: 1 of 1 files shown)


dannyhw:
weird that I never saw this before but I guess it should just be a regular dependency not dev dependency right?

duguyihou:
In storybookJS project, it is in devDependency as well. image

dannyhw:
hmm well its probably just prebundling it thats why, anyway I think this is fine to merge

dannyhw:
Thanks for the contribution 👍

PR #641, opened 2024-11-02
feat: store panel open state in storage

tlow92:
## Issue:

In my team we often don't use the addons panel actively, but it opens again everytime you refresh storybook in web.

## What I did

I reused the storage solution from `lastViewedStoryIds` which leverages `store2` and store the addons panel and sidebar  open state in storage.


packages/react-native-ui/src/hooks/useStoreState.ts
```diff
@@ -0,0 +1,12 @@
+import { useEffect, useState } from 'react';
+import store from 'store2';
```
dannyhw: I don't think this store actually works, I'm realising now that its in the lastViewed hook but I think its just an artifact from me copy/pasting a bit too freely from the web code 😅

We have the storage from the getStorybookUI thats required maybe we can pass that down to the ui

Edit:

Wait i guess if this is for web only that would work 🤔

tlow92: Yes it's only web, but I can take a look at the other storage solution and migrate to that so maybe we can drop store2 usage later alltogether.

dannyhw: yeah the thing is that we also want tablets and other wide screen devices to work so that would be ideal

dannyhw:
@tlow92 I had a go at adjusting it to use the user provided storage, let me know what you think of this solution

tlow92:
@dannyhw Only checked code right now, but looks good to me! 👍 This is much cleaner I think

Issue #645, opened 2024-11-08
storybook requires is invalid on windows as of 8.4.x


trajano:
**Describe the bug**

```
/* do not change this file, it is auto generated by storybook. */

import { start, updateView } from "@storybook/react-native";

import "@storybookaddon-ondevice-notes\register";
import "@storybookaddon-ondevice-controls\register";
import "@storybookaddon-ondevice-actions\register";
import "@storybookaddon-ondevice-backgrounds\register";
```
Note the missing `/` and the incorrect `\`

trajano:
I am thinking it's something to do with https://github.com/storybookjs/react-native/commit/467b40ba6dfec88eb77469c1d5850470fefebe55 as this changed generate.js last week

trajano:
The patch I have corrects the common.js which fixes the generated file but not during runtime

dannyhw:
Ah i know the problem i think, path.join is killing me

trajano:
I couldn't find the part where you do it at runtime but just adding `.posix` after `path` should be sufficient.

dannyhw:
If I put out a prerelease version are you able to test this easily?

dannyhw:
@trajano can you test `8.4.2-alpha.0`

dannyhw:
should be fixed now in 8.4.2

PR #648, opened 2024-11-08
fix: addon path generating on windows

dannyhw:
Issue: #645 

## How to test


packages/react-native/scripts/common.js
```diff
@@ -69,7 +69,7 @@ function resolveAddonFile(addon, file, extensions = ['js', 'mjs', 'ts']) {
 
   for (const ext of extensions) {
     try {
-      const filePath = path.join(addon, `${file}.${ext}`);
+      const filePath = `${addon}/${file}.${ext}`;
```
trajano, line 72: looks janky but I guess it should work.

trajano:
Thanks looks like it exported correctly can you add in the fix for the withSTorybook.d.ts

dannyhw:
@trajano added a fix in 8.4.2-alpha.1 for withStorybook can you give it a try?

packages/react-native/scripts/common.js
dannyhw: 😅 yeah I know but atleast I know what its doing haha

trajano:
oof no it's not backward compatible but it is cleaner.  I really hate the nuances of the default exports

trajano:
ChatGPTs's answer

Yes, Lord Archimedes, you can export the function both ways by doing the following:

```typescript
export function withStorybook() {
  // function implementation
}

export default withStorybook;
```

This allows `withStorybook` to be imported as either a named or default export.

dannyhw:
@trajano unfortunately I couldn't get that to work but I used this and I think its working now
```
export { withStorybook };

module.exports = withStorybook;

module.exports.withStorybook = withStorybook;
```

8.4.2-alpha.2

trajano:
weird that you have to go through that. you use tsup the same way I do too.

dannyhw:
if you use it like this you get the type info properly too

```
const { withStorybook } = require('@storybook/react-native/metro/withStorybook');

module.exports = withStorybook(defaultConfig);
```

trajano:
you know the weird part, you have two .js files one in `metro` and the other in `dist/metro`

trajano:
This types properly
```
const {
  withStorybook,
} = require('@storybook/react-native/metro/withStorybook');
```
But is not backwards compatible.

The way I test is to go to visual studio code on my metro.config.js and check if it type checks correctly

trajano:
Is it working for you though @dannyhw ?

dannyhw:
It works but the default export doesnt show the types for some reason.

And the extra file is actually a fallback for file based resolution for example when packageExports is off

trajano:
Okay, so far it's working for me with the `export { withStorybook }`  style on both WebStorm and VSCode.  TSC doesn't do any typechecks from what I can tell even with `// @ts-check`.  I verified by intentionally putting in `enabled: "yes"` which validates on TSC but not VSCode and WebStorm

dannyhw:
im going to just push the path fix for now this withstorybook thing is proving to be more complex than expected

dannyhw:
i will create another pr to try fix the with storybook thing soon

Issue #667, opened 2025-01-20
Slowness when opening stories list


fabioatcorreia:
**Describe the bug**
After upgrading @storybook/react-native to the latest version (8.4.4), the time to open the list of stories has degraded (takes around 2 seconds).  

**To Reproduce**
1. Open React Native Storybook app
2. Click on Stories explorer
3. Bottom sheet is visible after some delay

**Expected behavior**
The stories list should be visible without a major delay.

**Screenshots**

https://github.com/user-attachments/assets/b67e4168-6722-4095-96c2-448e5c19580e

**Additional context**
The project has quite a bit of stories files: 234.

dannyhw:
@fabioatcorreia which version are you upgrading from? is this happening on both android and ios or just android?

fabioatcorreia:
I'm upgrading from version `6.5.7`, which used a different layout (there was no bottom sheet).  

It also happens on iOS.

https://github.com/user-attachments/assets/823f00ae-24dc-41fc-a350-5f5e46c65b09

fabioatcorreia:
Just looked a bit through the code and it seems to be because the list is not virtualized.  
Since it has a lot of stories to load, it will try to render all of them before opening the bottom sheet.  

- [BottomSheetScrollView](https://github.com/storybookjs/react-native/blob/19b9557d6c5b085c286259a73fb51f07f5a7e956/packages/react-native-ui/src/MobileMenuDrawer.tsx#L68)
- [Tree](https://github.com/storybookjs/react-native/blob/19b9557d6c5b085c286259a73fb51f07f5a7e956/packages/react-native-ui/src/Tree.tsx#L358)

fabioatcorreia:
I'm not familiar with it, but it seems like the best approach would be to use `BottomSheetFlatList` instead of `BottomSheetScrollView` to render the stories list, but this means there needs to be a refactor around the current `Sidebar` and `Tree` implementation.

dannyhw:
yeah thats probably true, flatlist would be the obvious choice but it didn't seem to have issues in my testing so thats the only reason I didn't initially do that. 

I will need to reproduce this slowness in the example project to validate that it actually improves with a flatlist

I can take a look at this soon.

dannyhw:
@fabioatcorreia 8.5.2 is out now which should fix the slowness :)

PR #670, opened 2025-01-21
feat: flatlist for stories WIP

dannyhw:
Issue: #667

## How to test


dannyhw:
@fabioatcorreia is this along the lines of what you were suggesting? (work in progress still)

fabioatcorreia:
It isn't exactly what I had in mind, because I saw that `Sidebar` was also used for the Desktop version, which this approach seems to also affect (not sure if a problem to use `BottomSheetFlatList` outside a `BottomSheetModal`).  
If that works, go for it 👍

dannyhw:
Yeah I still need to work on this more, I will test web a bit later

dannyhw:
will potentially revisit in future if I can make a few more optimizations

PR #671, opened 2025-01-24
fix: improve loading bottom sheet by using non modal version

dannyhw:
Issue: #667 

## What I did

flatlist was making things slower so I tried a different approach, turns out the modal version of bottomsheet causes some lag since it has to render everything on demand.

## How to test


dannyhw:
@fabioatcorreia I'm trying something different since I found that the flatlist didn't have any reasonable perf impact (i think because of complex nested items)

I found that scrollview was actually working better (with 200 generated stories)

I moved the bottom sheet from a modal style to just a regular component and made some other adjustments that seem to have a bigger impact on perf.

Can you try `8.5.2-alpha.1` and see how it is for you?

fabioatcorreia:
It's looking great 🚀   

[storybook-improvement.webm](https://github.com/user-attachments/assets/0b91743c-a463-4c40-8194-b017bead96bf)

dannyhw:
made some optimizations for rerenders 8.5.2-alpha.3 should be smoother for changing stories etc

PR #674, opened 2025-01-25
chore: ui improvements

alcpereira:
Issue: No issue created

## What I did

Fixed some UI inconsistencies with current Storybook Web implementation.

There are a few mores I aim to fix, but I prefer to double check first that the intent is to be very similar to the Web version first.


dannyhw:
Hey, thanks for contributing 🙇‍♂️. Yes, the goal is to be similar to the web implementation where possible. In fact it's very much based on the web code.

This seems quite reasonable. Always happy to receive contributions to improve anything.

Unfortunately I'm unavailable for this whole week so im not able to check it right away. I will take a look as soon as I can.

Thanks again.

dannyhw:
for future ui changes it would be cool to have some screenshots showing before/after if possible

dannyhw (approved):

dannyhw:
LGTM 👍

PR #697, opened 2025-03-13
fix: use storyNames in tree if available

aniretaketea:
**Issue**: storyName property of stories is ignored in tree. There is no open issue, but there is a FIXME in code

## What I did
Fixed index object generation in `prepareStories`

## How to test
An example story with storyName was added in `examples/expo-example` under `UI/Sidebar/Tree`


dannyhw:
Hey @aniretaketea thanks so for much for your contribution, I haven't had time to test yet but seems good. Will test it out soon but I anticipate no issues with merging this in.

thanks again.

aniretaketea:
@dannyhw glad to be of help and give back! We actively use RN storybook in our project, thanks for maintaining and growing it ♥️

PR #698, opened 2025-03-15
feat: v9 init

dannyhw:
Issue:

## What I did

starts the process of updating to v9

~(note that right now only alpha.3 is working and not alpha.4)~

~errors in alpha 4 seem to relate to vitest/mocking features being added ad core project annotations
`code/core/src/test/preview.ts`, `code/core/src/preview-api/core-annotations.ts`~

~issues were resolved but we still get issues in jest for some reason~

Issues in jest are due to a patched version of testing library in core, this will be resolved upstream

## How to test


dannyhw:
This will be merged into next to not keep this branch getting too large since this already works

outstanding issues are

- storybook/test fn spy is not working for actions
- testing library user event issues in patched version from core

v8.6.2 code will be found in branch v8.6.0-stable

features intended for 8.6 should be implemented there

Issue #715, opened 2025-04-16
Enabling `composite` or `declaration` on tsconfig fails on type checking `storybook.requires`.


jvliwanag:
**Describe the bug**
Enabling `composite` or `declaration` on tsconfig fails on type checking storybook.requires` with the following error:

```
src/storybook.requires.ts:62:14 - error TS4023: Exported variable 'view' has or is using name 'Storage' from external module "node_modules/@storybook/react-native/dist/index" but cannot be named.

62 export const view = global.view;
```

**To Reproduce**
1. On a project with storybook react native, set `composite: true` on tsconfig.json.

**Expected behavior**
For `tsc` to succeed.

**Fix**

On the generated `storybook.requires`, add a type annotation:

```
export const view: ReturnType<typeof start> = global.view;
```

PR #716, opened 2025-04-16
fix: add type for view export (#715)

jvliwanag:
Issue: #715

## What I did

Add type for generated view export.


packages/react-native/scripts/generate.js
```diff
@@ -135,7 +135,7 @@ function generate({ configPath, absolute = false, useJs = false }) {
     updateView(global.view, annotations, normalizedStories, ${options});
   }
 
-  export const view = global.view;
+  export const view: ReturnType<typeof start> = global.view;
```
dannyhw, line 138: to support the js option what about something like this?

suggestion
  export const view${useJs ? '' : ': ReturnType<typeof start>'} = global.view;
```

Comment (dannyhw) [inline packages/react-native/scripts/generate.js]: To support the useJs option what about adding something like this?

```suggestion
  export const view${useJs ? '' : ': ReturnType<typeof start>'} = global.view;
```

jvliwanag: oops, missed useJs check. Fixed now

dannyhw:
exported the view type and used it directly instead of the return type from start, also exported a few more types that could have been causing issues.

PR #758, opened 2025-06-19
chore: improve generated storybook requires formatting

alcpereira:
## What I did

When bumping to v9 I noticed some small formatting issue on the generated file.
It will still be potentially conflicting with existing formatter configs, but we can improve some part of it.

## How to test

Run `generate.js` file


dannyhw:
Thank you 🙇‍♂️

PR #763, opened 2025-06-29
Fix: date shift bug in Date control component

ozgegurel:
## Issue : Date control displays wrong day on web - minor oversight in date formatting

The Date control component in `@storybook/ondevice-controls` had a date shift bug where the datetime picker would display the wrong day. For example, it would show "26.02.1983, 00:00" instead of the correct "25.02.1983, 00:00" due to an incorrect +1 offset applied to the `getDate()` method in the `webDateString` formatting.

datetime_picker

## What I did

- Fixed the date formatting logic in the `Date` control component
- Removed the incorrect +1 offset from `getDate()` in the `webDateString` formatting
- This ensures the `HTML` `datetime-local` input displays the correct date without any day shift

## How to test

1. **Install dependencies and build the project from the root directory:**
   ```sh
   yarn install
   yarn build
   ```
2. **Navigate to the example app and run it on WEB:**
   ```sh
   cd examples/expo-example
   yarn web
   ```
3. Open the `Date` control component in Storybook on web.
4. Set a specific date (e.g., February 25, 1983)
5. Verify that the datetime picker displays the correct date without any day shift
6. Test with different dates to ensure the fix works consistently

- **Does this need a new example in examples/expo-example?**  
  No, this is a bug fix for existing functionality and doesn't require new examples.

- **Does this need an update to the documentation?**  
  No, this fix doesn't change the public API or usage patterns, so documentation updates are not needed.


dannyhw (approved):
Good spot 😄, thanks for the contribution 🙇‍♂️

PR #768, opened 2025-07-09
feat: experiment with esm only build from core

dannyhw:
Issue:

## What I did

This PR is an **experimental branch** testing ESM-only builds from Storybook core.

### Context
To make v10 happen, we need to support a full ESM build of core. This PR tests integration with the ESM-only canary build (`0.0.0-pr-31819-sha-dcea463b`) from core Storybook.

### Key Changes
- Updated all Storybook dependencies to use the ESM-only canary build
- Migrated test infrastructure where needed (e.g., moved generate.test.js to Node.js test runner)
- Ensured compatibility with ESM modules throughout the React Native packages


dannyhw:
moving to #786

PR #815, opened 2025-11-29
feat: add storybookUIVisibility and layout parameter

dannyhw:
Issue: N/A

## What I did

Added two new story parameters for controlling UI behavior:

### `storybookUIVisibility`
Controls the initial visibility of the Storybook UI. Accepts `'visible'` or `'hidden'`. When set to `'hidden'`, the story starts in fullscreen mode with the UI hidden.

### `layout`
Controls the layout of the story container. Accepts `'fullscreen'`, `'centered'`, or `'padded'`:
- `'padded'` - adds 8px padding around the story
- `'centered'` - centers the story content
- `'fullscreen'` - no additional spacing

Changes made to:
- `packages/react-native-ui/src/Layout.tsx`
- `packages/react-native-ui-lite/src/Layout.tsx`
- `packages/react-native/src/components/StoryView/StoryView.tsx`
- Documentation updated

## How to test

1. Add `storybookUIVisibility: 'hidden'` to a story's parameters to have it start with UI hidden
2. Add `layout: 'centered'` to center your story content
3. Navigate between stories and verify the states update based on parameters

```tsx
export const MyStory = {
  parameters: {
    storybookUIVisibility: 'hidden', // 'visible' | 'hidden'
    layout: 'centered', // 'fullscreen' | 'centered' | 'padded'
  },
};
```

- Does this need an update to the documentation? Yes - done

### DIFF
--- a/docs/docs/intro/writing-stories.md
+++ b/docs/docs/intro/writing-stories.md
@@ -137,11 +137,11 @@ This parameter would instruct the backgrounds addon to reconfigure itself whenev
 
 React Native Storybook provides several built-in parameters to control the on-device UI behavior:
 
-| Parameter | Type | Description |
-|-----------|------|-------------|
-| `noSafeArea` | `boolean` | When `true`, removes the top safe area padding, allowing your story to render edge-to-edge |
-| `fullscreenDefaultValue` | `boolean` | When `true`, the story will start in fullscreen mode with the UI hidden |
-| `hideFullScreenButton` | `boolean` | When `true`, hides the fullscreen toggle button |
+| Parameter                | Type      | Description                                                                                |
+| ------------------------ | --------- | ------------------------------------------------------------------------------------------ |
+| `noSafeArea`             | `boolean` | When `true`, removes the top safe area padding, allowing your story to render edge-to-edge |
+| `fullscreenDefaultValue` | `boolean` | When `true`, the story will start in fullscreen mode with the UI hidden                    |
+| `hideFullScreenButton`   | `boolean` | When `true`, hides the fullscreen toggle button                                            |
 
 ```tsx
 // Button.stories.tsx
--- a/docs/docs/intro/writing-stories.md
+++ b/docs/docs/intro/writing-stories.md
@@ -133,6 +133,50 @@ export default meta;
 
 This parameter would instruct the backgrounds addon to reconfigure itself whenever a Button story is selected. Most addons are configured via a parameter-based API and can be influenced at a global, component and story level.
 
+#### UI-related parameters
+
+React Native Storybook provides several built-in parameters to control the on-device UI behavior:
+
+| Parameter | Type | Description |
+|-----------|------|-------------|
+| `noSafeArea` | `boolean` | When `true`, removes the top safe area padding, allowing your story to render edge-to-edge |
+| `fullscreenDefaultValue` | `boolean` | When `true`, the story will start in fullscreen mode with the UI hidden |
+| `hideFullScreenButton` | `boolean` | When `true`, hides the fullscreen toggle button |
+
+```tsx
+// Button.stories.tsx
+import type { Meta, StoryObj } from '@storybook/react-native';
+
+import { Button } from './Button';
+
+const meta: Meta<typeof Button> = {
+  component: Button,
+};
+
+export default meta;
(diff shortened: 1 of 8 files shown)


dannyhw:
@MarceloPrado let me know what you think :)

packages/react-native-ui-lite/src/Layout.tsx
```diff
@@ -102,6 +102,15 @@ export const Layout = ({
 
   const [uiHidden, setUiHidden] = useState(false);
 
+  useEffect(() => {
```
MarceloPrado: should we use `useLayoutEffect` instead? this way, it updates before the story is painted/rendered

docs/docs/intro/writing-stories.md
```diff
@@ -133,6 +133,50 @@ export default meta;
...
 This parameter would instruct the backgrounds addon to reconfigure itself whenever a Button story is selected. Most addons are configured via a parameter-based API and can be influenced at a global, component and story level.
 
+#### UI-related parameters
+
+React Native Storybook provides several built-in parameters to control the on-device UI behavior:
+
+| Parameter                | Type      | Description                                                                                |
+| ------------------------ | --------- | ------------------------------------------------------------------------------------------ |
+| `noSafeArea`             | `boolean` | When `true`, removes the top safe area padding, allowing your story to render edge-to-edge |
+| `fullscreenDefaultValue` | `boolean` | When `true`, the story will start in fullscreen mode with the UI hidden                    |
```
MarceloPrado: I originally thought about a different API:
```tsx
declare module "storybook/internal/types" {
  declare interface Parameters {
    /**
     * The layout to use for the story. 
     * 
     * By default, Storybook renders your story in a container that also renders
     * the addon pannel. This can be an issue when testing components that rely
     * on the bottom inset or the keyboard position. This can be resolved by
     * using the `fullscreen` layout.
     * 
     * - `default` renders the story in a container with the addon pannel
     * - `fullscreen` renders the story without any container
     * 
     * @default "default"
     */
    layout?: "fullscreen" | "default";
  }
}
```

However, I didn't realize you already supported a few other params 😄

Also, regardless of the API you decide to use, one thing that would be very useful/handy to document is how to get strong type-safety in your `parameters` field. The snippet above enhances the TS declaration with such capability. I don't know if this should be baked-in by Storybook or not, though. I personally believe that if Storybook already supports some parameters, it would be handy to display them in IntelliSense/TS. It's great when the API is self-explanatory.

dannyhw: Yeah I suppose it's possible but layout: fullscreen has a slightly different meaning for storybook like centred/fullscreen/padding controls how the story is placed within the canvas (This is something that I do plan to support too). Whereas this is more about hiding the storybook ui. Maybe it should be called hideUiDefault or something instead 

Regarding types thats actually something csf4 factories is addressing and its probably better addressed through that since parameter types is a long standing problem really. Though I could export a type that can be extended user side to make it easier to use today.

I will revisit csf factories soon to make sure they are working in React Native.

dannyhw: Also an outstanding issue is supporting globals like on the web, I tried a few times but theres some incompatibility somewhere I haven't solved yet.

packages/react-native-ui-lite/src/Layout.tsx
dannyhw: Hmm good question, I can try that.

docs/docs/intro/writing-stories.md
MarceloPrado: yep, that makes sense! You have a lot more context on the storybook terms than I do + the direction things are going, so appreciate the push back 😄

What do you think of making this more explicit though vs. a boolean? For instance: `storybookUi: "hidden" | "expanded"`

dannyhw: Hmm yeah something like that can work, let me think on it.

Thanks for your input 🙏

dannyhw:
ended up adding the layout parameter as well

Heres what I've landed on

```
layout: 'padded', // fullscreen, centered, padded
storybookUIVisibility: 'visible', // visible, hidden
```

dannyhw:
@MarceloPrado you can try it now in 10.1.1

PR #823, opened 2025-12-07
feat:  add RN_GET_INDEX event and auto host detection

dannyhw:
### Auto WebSocket Host Detection
- Added `websockets: 'auto'` option in metro config that automatically detects the local IP address
- in generate --host auto or host: 'auto' will produce the same output as websockets auto
- The `getLocalIPAddress()` function finds the machine's IPv4 address from network interfaces
- Eliminates the need to manually configure host/port for device connections

### RN_GET_INDEX Event
- Added `RN_GET_INDEX` channel event that clients can send to request the story index
- Server responds with `RN_INDEX_RESPONSE` containing the current story index
- Enables external tooling (VS Code extension, dev tools) to query available stories

### Build-time WebSocket Config Injection
- New `STORYBOOK_WEBSOCKET` global variable injected into `storybook.requires` at build time
- Contains `{ host, port }` configuration determined during metro bundling
- `View.tsx` now reads from this global when `host`/`port` params are not explicitly provided

### Other Changes
- Added `--host` and `--port` CLI arguments to the generate script
- Added WebSocket ping interval (10s) to keep connections alive
- Standardized on `globalThis` instead of `global`
- Added test script (`scripts/test-ws.ts`) for testing WebSocket connection
- added /node entrypoint with build index and server channel exports
- create server channel is the rn storybook server exported for tooling use
- build index provides story list as a sync command in node

## How to test

1. In metro config, set `websockets: 'auto'`:
```js
module.exports = withStorybook(defaultConfig, {
  enabled: true,
  websockets: 'auto',
});
```

2. Run the app on a physical device - it should auto-connect to the WebSocket server

3. Test the RN_GET_INDEX event using the test script:
```bash
cd examples/expo-example
npx ts-node scripts/test-ws.ts <host> <port>
```


packages/react-native/src/metro/channelServer.ts
```diff
@@ -0,0 +1,177 @@
...
+  configPath,
+}: ChannelServerOptions): WebSocketServer {
+  const httpServer = createServer(async (req: IncomingMessage, res: ServerResponse) => {
+    if (req.method === 'OPTIONS') {
+      res.writeHead(204);
+      res.end();
+      return;
+    }
+
+    if (req.method === 'GET' && req.url === '/index.json') {
```
dannyhw, line 50: instead I will do two endpoints, one for ws based and another for file based

problem with file based is we can't easily load the story sort from preview since react native's flow breaks the parsing of the file.

dannyhw: or maybe i will just try to mock out react native imports for this scenario

PR #826, opened 2025-12-15
feat: CSF Factories (CSF Next) support

dannyhw:
## Summary

10.3.0-canary-20260301145716

Adds support for [CSF Next](https://storybook.js.org/docs/api/csf/csf-next), the new Storybook configuration and story authoring format available experimentally in Storybook 10.

CSF Next replaces the traditional `export default meta` / `export const Story` pattern with a `definePreview()` / `defineMain()` / `meta.story()` API that provides full type safety, composability, and better IDE support.

### Core changes

- **`definePreview()` support** — Re-exports `definePreview` from `@storybook/react` and adds `resolveAnnotations()` in `Start.tsx` to unwrap the opaque `Preview` object that `definePreview()` produces. We extract `.input` (the user's plain config) rather than using `.composed`, which includes web-specific core annotations that depend on a browser environment not available on-device.
- **`defineMain()` support** — Adds a typed `defineMain()` helper exported from `@storybook/react-native/node` for type-safe main config.
- **`StorybookConfig` type refinement** — Moved to `types/config.ts`, now uses `Preset` type for addons instead of loose string/object union.
- **Improved type annotations** — Replaces `any[]` with `ModuleExports[]` and `unknown` throughout `Start.tsx`.

### Examples and tests

- Adds `.factories.stories.tsx` variants for all existing example stories demonstrating the factories API
- Adds `.factories.test.tsx` portable story tests for factories stories
- Adds `.rnstorybook-nofactories/` config directory to test both config styles side-by-side
- Updates `metro.config.js` and `App.tsx` to support switching between factories and non-factories configs

### How the factories API looks

**preview.tsx (before):**
```tsx
import type { Preview } from '@storybook/react-native';

const preview: Preview = {
  parameters: { /* ... */ },
};
export default preview;
```

**preview.tsx (after):**
```tsx
import { definePreview } from '@storybook/react-native';

export default definePreview({
  addons: [],
  parameters: { /* ... */ },
});
```

**Stories (before):**
```tsx
const meta = { component: Button } satisfies Meta<typeof Button>;
export default meta;
export const Primary: StoryObj<typeof meta> = { args: { label: 'Click' } };
```

**Stories (after):**
```tsx
const meta = config.meta({ component: Button });
export const Primary = meta.story({ args: { label: 'Click' } });
```

## Test plan

- [ ] `pnpm build` completes successfully
- [ ] `pnpm test` passes — both factories and non-factories story tests
- [ ] Run expo example with factories config — stories render correctly
- [ ] Run expo example with non-factories config — stories render correctly
- [ ] `definePreview()` config is properly unwrapped (parameters, decorators, etc. are applied)
- [ ] `defineMain()` provides type checking for main config


dannyhw:
current issues:
anything that tries to read meta/preview can sometimes get nothing because of how we now have .input and .composed instead of plain objects

dannyhw:
update: added a react native define preview so we get compose preview, story etc without issues and more inline with storybook/react

Issue #822, opened 2025-12-07
Expo + Storybook Bundler Issue metro config


ian-etgo:
When adding storybook to the metro.config.json

This error is seen:

<img width="636" height="583" alt="Image" src="https://github.com/user-attachments/assets/15779c02-b07a-4f66-8344-61dd23555cf1" />

Expo:     "expo": "^54.0.27",
Storybook version:     "storybook": "^10.1.4",

```
const { getDefaultConfig } = require('expo/metro-config');
const { withNativeWind } = require('nativewind/metro');
const {
  withStorybook,
} = require('@storybook/react-native/metro/withStorybook');

const config = getDefaultConfig(__dirname);
const { transformer, resolver } = config;

const transformedResolvedConfig = {
  transformer: {
    ...transformer,
    babelTransformerPath: require.resolve('react-native-svg-transformer'),
  },

  resolver: {
    ...resolver,
    assetExts: resolver.assetExts.filter(ext => ext !== 'svg'),
    sourceExts: [...resolver.sourceExts, 'ts', 'tsx', 'js', 'jsx', 'svg'],
  },
};

const nativeWindConfig = withNativeWind(transformedResolvedConfig, {
  input: './global.css',
});

//this must be commented out
// const storybookconfig =  withStorybook(nativeWindConfig);

module.exports = nativeWindConfig;
```

dannyhw:
Can you provide a reproduction? theres not much to go off of here unfortunately.

dannyhw:
no reproduction and no response means theres not much I can do here, closing for now. If you can provide more information I could investigate

mzaien:
it is node version, my eas config was using node 22.x now it uses 22.4 and all good

PR #832, opened 2026-01-11
fix: replace @ts-ignore with Metro-specific type definitions

YevheniiKotyrlo:
## Summary

Removes `@ts-ignore` comment from generated `storybook.requires.ts` by providing proper Metro type definitions.

**Previous approach (rejected):** Added `@types/webpack-env` as dependency.

**New approach (per maintainer feedback):** Custom Metro type definitions following the same pattern used by Expo.

## Background Research

Metro's `require.context` API intentionally matches webpack's API by design ([Metro PR #822](https://github.com/facebook/metro/pull/822)), but the runtime implementations differ. Rather than using webpack types, this PR provides Metro-specific types.

**How Expo handles this:**
- Expo manually maintains Metro types at [`packages/expo/types/metro-require.d.ts`](https://github.com/expo/expo/blob/main/packages/expo/types/metro-require.d.ts)
- Created by Mark Lawlor in [PR #24255](https://github.com/expo/expo/pull/24255) (Sept 2023)
- Uses `__MetroModuleApi` namespace to avoid conflicts with other type definitions
- Based on webpack-env types but customized for Metro's implementation

**Why not auto-convert from Flow?**
- Metro is written in Flow, not TypeScript
- Automated Flow-to-TypeScript converters fail on Metro's complex type definitions
- No one in the ecosystem (Expo, React Native) auto-converts these types
- Manual types with clear source references is the standard approach

## Changes

1. **`metro-env.d.ts`** - New file with Metro type definitions
   - `__MetroModuleApi.RequireContext` - require.context() return type
   - `__MetroModuleApi.RequireFunction` - require.context() method signature
   - `__MetroModuleApi.Hot` - module.hot HMR API
   - Extends `NodeJS.Require` and `NodeModule` interfaces

2. **`package.json`**
   - Removed `@types/webpack-env` dependency
   - Added `"./metro-env": "./metro-env.d.ts"` export

3. **`scripts/generate.js`**
   - Adds `/// <reference types="@storybook/react-native/metro-env" />` to generated TypeScript files

## References

- [Expo's metro-require.d.ts](https://github.com/expo/expo/blob/main/packages/expo/types/metro-require.d.ts)
- [Metro PR #822 - require.context implementation](https://github.com/facebook/metro/pull/822)
- [Metro HMR Flow types](https://github.com/facebook/metro/blob/main/packages/metro-runtime/src/polyfills/require.js.flow)
- [Evan Bacon's Metro Context Demo](https://github.com/EvanBacon/Metro-Context-Modules-Demo)

## Test Plan

- [x] `yarn test:generate` passes
- [x] Generated files include proper type reference
- [x] No `@ts-ignore` needed in generated code


dannyhw:
Hey thanks for this PR, however I wouldn't want to add this dependency personally.

I understand the intention behind this pr, however I don't think this is the right fix since the metro implementation doesn't match the webpack one and we shouldn't rely on any dependency to webpack in my opinion.

If anything we should copy the type definitions from expo here
https://github.com/expo/expo/blob/main/packages/expo/types/metro-require.d.ts

However its also low priority in my opinion since this ts-ignore exists in a generated file that can be ignored in eslint

YevheniiKotyrlo:
Hey @dannyhw, I've rebased this onto the latest `next` and addressed your feedback from the previous iteration:

- Removed `@types/webpack-env` entirely
- Added custom `metro-env.d.ts` with Metro-specific types in the `__MetroModuleApi` namespace, following the same approach Expo uses in [their metro-require.d.ts](https://github.com/expo/expo/blob/main/packages/expo/types/metro-require.d.ts)
- The generated `storybook.requires.ts` now uses `/// <reference types="@storybook/react-native/metro-env" />` instead of `@ts-ignore`
- Added a changeset

No new dependencies added. Would you be able to take another look when you get a chance?

dannyhw:
Hey, thanks for that. I'm just wondering if you could clarify  whats the pain point that this solves? Does this tsignore break something for you currently?

Just so i can understand better where this is coming from

YevheniiKotyrlo:
The main pain point is that the generated `storybook.requires.ts` can't participate in the project's type-checking and linting pipeline.

Our project runs `strict: true` TypeScript with `strictTypeChecked` ESLint rules. The `@ts-ignore` comments in the generated file mean we have to explicitly exclude it from our ESLint config:

```typescript
// eslint.config.ts — ignores
'**/storybook/storybook.requires.ts',
```

This creates a blind spot — if the generated code has an actual type error beyond the missing `require.context`/`module.hot` types, it's invisible to our CI. The `@ts-ignore` suppresses all errors on the line, not just the missing type.

With proper Metro types via `metro-env.d.ts`, the generated file can be linted and type-checked like everything else — no exclusions needed.

dannyhw:
you actually should ignore the requires file from eslint, its a generated file and its not possible to satisfy every project's lint configuration.

I think its valid to aim to remove ts ignore but i don't think it will fix linting problems.

YevheniiKotyrlo:
That's a good point — and you're right that ignoring it from ESLint is the practical approach. The challenge though is that `storybook.requires.ts` gets regenerated by `generate()` on every `expo start` via the `withStorybook()` Metro wrapper. So even if a user removes the `@ts-ignore` manually, the next dev server start overwrites the file and brings them back.

The only way to prevent that currently is to patch `@storybook/react-native`'s `scripts/generate.js` — which is what we've been doing on our end. This PR removes the need for that kind of workaround by having `generate()` emit a `/// <reference types>` directive instead of `@ts-ignore`, backed by proper Metro types in `metro-env.d.ts`.

YevheniiKotyrlo:
Let me know if you have any concerns about the approach or if there's anything else you'd like me to test or adjust before this can be merged. Happy to make changes if needed!

dannyhw:
Ok thanks, I will take a proper look tomorrow. Sorry that its been stuck for a while.

dannyhw:
Doesn't appear to work for module.hot

```
> expo-example@10.2.2-alpha.5 check /Users/danielwilliams/Developer/storybook/react-native-storybook/examples/expo-example
> tsc --noEmit

.rnstorybook/storybook.requires.ts:66:9 - error TS2339: Property 'hot' does not exist on type 'Module'.

66 module?.hot?.accept?.();
           ~~~


Found 1 error in .rnstorybook/storybook.requires.ts:66
```

applied a small fix

dannyhw:
available to try on `10.3.0-canary-20260220230836`

dannyhw:
thanks for your work and patience @YevheniiKotyrlo

YevheniiKotyrlo:
> thanks for your work and patience @YevheniiKotyrlo

Thank you for your hard work!

Issue #850, opened 2026-02-25
AAPT2 compile error for react-native-logo.png


andre-krueger:
**Describe the bug**
The mentioned file is causing compile errors, as they are MS Windows icon resources, while their extension is PNG.

**To Reproduce**
Build a React Native app with Storybook setup in release mode, with AAPT2 enabled. 

**Expected behavior**
The app should build successfully.

**Code snippets**
```
file .../android/app/build/generated/res/.../drawable-mdpi/node_modules_storybook_reactnativeuicommon_src_assets_reactnativelogo.png 

.../drawable-mdpi/node_modules_storybook_reactnativeuicommon_src_assets_reactnativelogo.png: MS Windows icon resource - 4 icons, 16x16, 32 bits/pixel, 24x24, 32 bits/pixel
```

**System:**
```
│  System:
│  OS: macOS 26.2
│  CPU: (12) arm64 Apple M2 Pro
│  Shell: 5.9 - /bin/zsh
│  Binaries:
│  Node: 23.3.0 - /Users/andre/.asdf/installs/nodejs/23.3.0/bin/node
│  Yarn: 4.7.0 - /Users/andre/.asdf/installs/nodejs/23.3.0/bin/yarn <----- active
│  npm: 10.9.0 - /Users/andre/.asdf/plugins/nodejs/shims/npm
│  Browsers:
│  Chrome: 145.0.7632.116
│  Firefox: 141.0
│  Safari: 26.2
│  npmPackages:
│  @storybook/addon-ondevice-actions: 10.2.3 => 10.2.3
│  @storybook/addon-ondevice-controls: 10.2.3 => 10.2.3
│  @storybook/react: 10.2.10 => 10.2.10
│  @storybook/react-native: 10.2.3 => 10.2.3
│  storybook: 10.2.10 => 10.2.10
```

dannyhw:
Im not sure i understand what the issue is? What png is that? I don't believe any png is shipped with the package.

andre-krueger:
It's this file: https://github.com/storybookjs/react-native/blob/next/packages/react-native-ui/src/assets/react-native-logo.png
It's in all ui-packages.

andre-krueger:
This issue seems to be caused by the Lite UI changes (which I don't use), as I didn't have any AAPT issues with RN Storybook V8.

It should be fixed by re-exporting these images as PNGs. I tried patching them with Yarn, but the patches can't be applied for some reason.

dannyhw:
[...]

PR #851, opened 2026-02-25
fix: Re-export logo as PNG

andre-krueger:
Issue: https://github.com/storybookjs/react-native/issues/850

## What I did
Re-export the logo to PNG
## How to test
Build an Android release build with AAPT2 enabled. The build should succeed.


dannyhw:
preferably we would add this to npm ignore because its not intended to be included in the bundle

PR #876, opened 2026-04-11
Fix os and tty modules being emptied with Expo API Routes

Nezz:
Issue: `withStorybook` replaces os and tty with empty modules on all platforms, breaking Expo API Routes

## What I did

`withStorybook` unconditionally resolved os and tty to `{ type: "empty" }` as a workaround. This breaks Expo API Routes (which bundle for the web platform) when any server-side dependency needs real Node built-ins like `os.hostname()`.

Changes:
- Enabled path: Scoped the os/tty empty-module override to native platforms only (`platform !== 'web'`), so web/server bundles get the real Node modules.
- Disabled path: Removed the os/tty override entirely. When Storybook is disabled, all `storybook/@storybook` imports are already emptied, so the instrumentor is never loaded and its os/tty imports are never reached.
- Added tests covering both behaviors.

## How to test

1. Set up an Expo project with API Routes and `@storybook/react-native`
2. In an API route, use any library that calls `require("os")` (e.g. `os.hostname()`)
3. Verify the API route works without `os.hostname is not a function` errors
4. Verify Storybook still loads correctly on iOS/Android (the os/tty override still applies on native platforms)


packages/react-native/src/metro/withStorybook.ts
```diff
@@ -166,13 +166,6 @@ export function withStorybook(
             };
           }
 
-          // workaround for node imports in instrumentor.cjs
-          if (moduleName === 'tty' || moduleName === 'os') {
-            return {
-              type: 'empty',
-            };
-          }
```
Nezz, line 174: The intention behind doing this when storybook is disabled wasn't clear to me, but this could be re-added if needed with the web check

dannyhw: Its because I had some reports of these packages getting pulled in somehow even when storybook was disabled

dannyhw:
Makes sense, thanks for the pr 🙏

packages/react-native/src/metro/withStorybook.ts
Nezz: Oh wow, not sure how that could happen 😅

dannyhw: Yeah i don't know, just metro things 😅

dannyhw (approved):

PR #894, opened 2026-05-29
feat: enhance Storybook server configuration with environment variable support

ndelangen:
feat: respect `STORYBOOK_SERVER` env variable in legacy metro and repack config wrappers

## What I did

The `STORYBOOK_SERVER=false` env variable was already supported in the newer entrypoint-swapping `withStorybook` path, but was not being respected in the legacy `metro/withStorybook` and `repack/withStorybook` config wrappers. This meant customers using the legacy setup who set `STORYBOOK_SERVER=false` would still have the channel server attempt to start — causing SSL errors when no cert/key was configured:

```
Error: [Storybook] Secure channel server requires both `ssl.key` and `ssl.cert`.
```

Added `envVariableToBoolean` support for `STORYBOOK_SERVER` in both legacy wrappers, defaulting to `true` for full backwards compatibility. The channel server is now skipped entirely when `STORYBOOK_SERVER=false`, rather than attempting to start and failing on SSL validation.

## How to test

- Set `STORYBOOK_SERVER=false` in your environment when using the `metro/withStorybook` config wrapper — the channel server should not start, and no SSL error should appear.
- Verify existing setups without `STORYBOOK_SERVER` set continue to work as before (default is `true`).

Validated via canary release `10.4.5-canary-20260529144515` — multiple customers confirmed the fix resolves the SSL error without needing to remove env vars.


packages/react-native/src/metro/withStorybook.ts
```diff
@@ -144,6 +144,7 @@ export function withStorybook(
   } = options;
 
   const disableTelemetry = optionalEnvToBoolean(process.env.STORYBOOK_DISABLE_TELEMETRY);
+  const server = envVariableToBoolean(process.env.STORYBOOK_SERVER, true);
```
dannyhw, line 147: why wouldnt we use existing channelWebsocketsEnabled?

ndelangen: That's an option that's set/can be set at runtime. We'll want to have an env-variable based override for chromatic

ndelangen: The reasoning is to make the old setup easier to set up with Chromatic as well.

PR #895, opened 2026-06-02
fix: mark Storybook generated View import as type-only

saseungmin:
Issue:

Generated `storybook.requires.ts` imports `View` as a runtime binding even though it is only used as a TypeScript type. This causes TypeScript errors in projects using `verbatimModuleSyntax: true`.

## What I did

Marked the generated `View` import as type-only:

```ts
import { start, updateView, type View, type Features } from '@storybook/react-native';
```

Updated the generator snapshot tests to match the new output.

## How to test

Run:

```bash
pnpm --filter @storybook/tests test
```

I also verified the React Native package type check:

```bash
pnpm --filter @storybook/react-native check:types
```

    - No. This only changes the generated TypeScript import syntax and does not add new behavior.

    - No. This is a compatibility fix for verbatimModuleSyntax and does not change user-facing APIs or setup steps.


dannyhw (approved):

dannyhw:
Thanks!

PR #904, opened 2026-07-02
fix: use unmapped args for on-device controls

adam-sajko:
Issue: native side of storybookjs/storybook#19130

Related (Core): https://github.com/storybookjs/storybook/pull/35360

## What I did

Fixed mapped `select` controls and conditional `if` on device.

`useArgs` was reading `story.args`, which are the mapped values. Controls need the raw option keys (`unmappedArgs`), same as web. Because of that, a mapped `select` couldn't match its value back to an option, and any `if` that points at a mapped arg was compared against the mapped value instead of the key, so it toggled wrong.

Now it reads `unmappedArgs` (falls back to `args`) for the initial state, on story change, and on `STORY_ARGS_UPDATED`. The event payload is mapped, so I re-read the story to get the unmapped values. Same behavior as web now.

## How to test

Story with a mapped `select` and an arg that depends on it:

```ts
argTypes: {
  icon: {
    control: { type: 'select' },
    options: ['none', 'star'],
    mapping: { none: undefined, star: <StarIcon /> },
  },
  iconPosition: {
    if: { arg: 'icon', eq: 'star' },
    control: { type: 'radio' },
    options: ['left', 'right'],
  },
}
```

Set `icon` to `star` on device:

- Before: `select` doesn't reflect `star`, `iconPosition` doesn't show/apply.
- After: `star` is selected, `iconPosition` shows and gets passed to the component.

- Does this need a new example in examples/expo-example? No, any story mixing `mapping` with an `if` reproduces it.
- Does this need an update to the documentation? No, no API change, just restores web parity.


packages/ondevice-controls/src/hooks.ts
```diff
@@ -6,6 +6,9 @@ import {
   STORY_ARGS_UPDATED,
 } from 'storybook/internal/core-events';
 
+const getControlArgs = (story: StoryContext): Args =>
+  (story as StoryContext & { unmappedArgs?: Args }).unmappedArgs ?? story.args;
```
dannyhw: I wonder why this story context type isn't covering this unmappedArgs property 🤔

and perhaps we can have a `PreparedStoryContext` type alias for this?

packages/ondevice-controls/src/hooks.ts
```diff
@@ -15,20 +18,24 @@ export const useArgs = (
...
+    }
     // Sync the args up with the initial args of the story, since the story ID
     // must have changed for this effect to run.
-    setArgs(initialArgs);
+    setArgs(getControlArgs(currentStory));
     const cb = (changed: { storyId: string; args: Args }) => {
       if (changed.storyId === storyId) {
-        setArgs(changed.args);
+        const updatedStory: StoryContext | undefined = storyStore.fromId(storyId);
+        setArgs(updatedStory ? getControlArgs(updatedStory) : changed.args);
```
dannyhw: is it necessary here to use getControlargs in the callback here? are the args from STORY_ARGS_UPDATED not the unmapped ones?

dannyhw: adjusted

packages/ondevice-controls/src/hooks.ts
dannyhw: adjusted

PR #919, opened 2026-08-31
feat: CSF Factories (CSF Next) support

kasperpeulen:
Issue: supersedes #826 (same branch content, brought up to date with `next`)

## What I did

Fresh PR of @dannyhw's CSF Factories (CSF Next) work from #826, which had drifted ~6 months behind `next` (74 commits) and accumulated merge conflicts. All 24 of Danny's original commits are preserved; on top of them sit one merge commit resolving the conflicts against current `next` and one follow-up chore commit.

Danny's work (see #826 for the full description):

- RN-specific `definePreview()` wrapping `storybook/internal/csf`, injecting the RN entry-preview annotations and render function, with compat shims so `prepareStories` can read `meta.title`/`includeStories`/`excludeStories` and stories expose a lazy `Component` and a `play` that stays `undefined` when unset
- Typed `defineMain()` exported from `@storybook/react-native/node`
- `.factories.stories.tsx` variants of all example stories plus portable-story tests, and a `.rnstorybook-nofactories/` config so both CSF styles run side by side

Merge resolution highlights:

- `Start.tsx` / `preview.ts`: took `next`'s refactored versions (extracted `getReactNativeProjectAnnotations`, `StoryEntry` type, web workarounds); the factories logic lives in `definePreview.tsx` / `entry-preview.ts`
- `index.ts`: kept the `types/config.ts` split and `definePreview` exports, with `StorybookConfig` updated to `next`'s newer shape (`deviceAddons`, deprecated `addons`) — this drops the `addons: Preset[]` refinement from #826 in favor of `next`'s deliberate `Addon` union + `deviceAddons` split
- `prepareStories.ts`: `next`'s TS6-friendly typing plus the factories-aware story-name resolution and `typeof story.play === 'function'` detection
- `.rnstorybook-nofactories/main.ts`: migrated from the deprecated `addons` field to `deviceAddons` and regenerated its stale `storybook.requires.ts`

## How to test

```sh
pnpm install
pnpm build
pnpm check:types   # passes
pnpm lint          # passes
pnpm format:check  # passes
pnpm test          # passes (one pre-existing flake: Actions.test.tsx times out under parallel load, passes in isolation)
```

Verified against the expo example running via `expo start` (web target), with the factories config as the default `.rnstorybook`:

- all `.factories` stories are indexed and render (Actions, Text Input, Deep Controls, Backgrounds, Notes, Hidden Controls, plus the Factories Edge Cases group)
- ondevice-actions logs `onPress` from a factories story
- ondevice-controls binds to a factories story's args and live-edits them (canvas updates immediately)
- `sb-rn-get-stories` generates correctly for both the factories and nofactories configs

Not yet covered here: an on-device (simulator) run — the jest portable-story tests and the web target exercise the same composition path, but a native run is worth doing before merging. One observation, present for both classic and factories stories (so unrelated to this PR): the controls panel doesn't refresh when switching stories and only binds correctly after a reload.

- Does this need a new example in examples/expo-example? Already included (factories story variants + nofactories config).
- Does this need an update to the documentation? Yes, once the API is considered stable enough to advertise — CSF Next is still a preview feature in core.


dannyhw:
👀

dannyhw:
I think theres still the question of what we should do for users that have `@storybook/react-native-web-vite` and react native storybook together. Which preview do they import from and should we provide a suggested way to alias?

ndelangen:
@dannyhw we agree with you on the dual-install question (https://github.com/storybookjs/react-native/pull/919#issuecomment-5557421180).

For projects that run both `@storybook/react-native` and `@storybook/react-native-web-vite` against the same story files, CSF4 stories should keep the documented import:

```ts
import preview from '#.storybook/preview';
```

Metro should alias that to `{configPath}/preview` (default `.rnstorybook/preview`) inside `withStorybook`. Vite already loads `.storybook/preview` for real, so web does not need a special case.

That is the direction we want, rather than duplicating stories or teaching a native-only relative import in shared files.

Before we treat it as the public pattern, we will spike two things:

1. Metro actually resolving `#.storybook/preview` via the existing `withStorybook` `resolveRequest` hook.
2. TypeScript: a single `package.json` `"imports"` map still types that specifier as `.storybook/preview`. We need to see whether that is acceptable for dual apps, or whether we need two TS projects.

Until that spike lands, we would rather not freeze `import preview from '../../.rnstorybook/preview'` in the factory examples as the recommended API. Native-only `definePreview` in this PR is still useful; the story import is the part we do not want to lock yet.

Does that match what you had in mind for the alias?

dannyhw:
@ndelangen that seems reasonable. The part that I'm a bit unsure on is the typescript part, basically you're saying if we add that imports mapping to package.json typescript also uses that to give the right types? The user should then be conscious to not let the 2 preview files drift so the types remain accurate?

Will we provide that package imports mapping through migration/init script?

It seems like it also creates 2 scenarios, you have the react native only setup where you would want it to have the mapping be `.rnstorybook/preview` and then for web/rn projects it could be `.storybook/preview`? Maybe its simpler to have it reference `.rnstorybook/preview` in both cases?

dannyhw:
Also I'm only hesitant to merge because I think it could maybe be smart to delay this change for v11, and also now knowing that we want to avoid locking in that `../../../` approach

maybe there can be a pr against this one that shows the `#.storybook/preview` approach working which we can use to validate the idea?

ndelangen:
@dannyhw the mapping in `package.json` does not pick native types. It only says which **file** `#.storybook/preview` is.

The CSF4 `#*` map turns that specifier into `.storybook/preview`. TypeScript then types the web preview. Metro would too, unless we rewrite at resolve time (what we want for dual apps).

On drift: you’re right that it matters for dual apps if TypeScript only sees the web preview. Don’t try to keep the two `definePreview` files identical (addons and render stay different). Extract the overlap (shared `parameters`, `tags`, etc.) into one module both previews compose. Then the story-facing types stay accurate without people copying between files.

Init/migration should follow the topology:

- **RN only** (no `.storybook`): point the import at `.rnstorybook/preview` (either `#.rnstorybook/preview`, or map `#.storybook/preview` → `.rnstorybook/preview` so the documented specifier still works). Metro needs no rewrite.
- **RN + RNW:** keep `#.storybook/preview` in the shared story. Vite uses `.storybook/preview`. Metro (`withStorybook`) rewrites to `{configPath}/preview`. Native TypeScript still needs a second mapping (native `tsconfig` `paths`, or `"customConditions": ["react-native"]`).

I would not have every story import `.rnstorybook/preview`. That makes Vite/RNW the special case, and CSF4 already documents `#.storybook/preview`. Metro is the odd resolver; the specifier should stay the documented one.

ndelangen:
@dannyhw the stacked PR you asked for is here: https://github.com/storybookjs/react-native/pull/926

`withStorybook` rewrites `#.storybook/preview` to `{configPath}/preview`, and the expo CSF4 stories now use that specifier instead of a relative `.rnstorybook` import.

